PR #5, opened 2016-01-23
Updated trivia comparison to be more lenient

micmorris:


micmorris:
Not sure why the diff didn't actually work...

Kwoth:
very nice

PR #12, opened 2016-01-25
Added new command and add new command to read me

ZR2:
Added a color command so admins with the role manageRoles can set colors
took me a long time to make hope you like it


Kwoth:
Looks great, I will modify it a bit tho.

PR #21, opened 2016-02-14
Added the 3 other osu modes to search

ZR2:


Kwoth:
we will do this some other way

PR #48, opened 2016-02-18
Tagdef.com search added

Blaubeerwald:
Added hashtag lookup by modifying cuken's urban dictionary lookup. Sorry new to Github, so this request maybe is a bit messed up.

### DIFF
--- a/Searches.cs
+++ b/Searches.cs
@@ -1,0 +1,260 @@
+﻿using System;
+using Discord.Modules;
+using System.Net;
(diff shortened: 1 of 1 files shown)


Kwoth:
Hey, thanks

I have copied the ~# command from this PR and added it manually, since git diff didn't work for your PR.

Make sure to fetch before making a PR next time. (only thing covered in green should be your added piece of code, not the whole file)

PR #54, opened 2016-02-24
Update Searches.cs

rob5251:
Changed fail message for manga lookup - fail message returned "anime" instead of "manga"


Kwoth:
;o thx

rob5251:
No problem, I can't code; but at least I can proof read. :)

PR #150, opened 2016-03-25
Cleaner code

tg44:
So I started to clean and reorganize the Gambling module.

What I done:
- in the Module folder make a subfolder to a module and its assets (Gambling)
- moved all the commands and the module to that folder
- renamed the module (Gambling->GamblingModule)
- fixed the namespaces
- in the controllers extract all the inline functions
- renamed the Tasks to refer what they do (DoTask -> ReshuffleTask)
- reorder the functions to the Constructor, Init, private functions order
- cleaned unused imports, reorder imports, autoformat code
- moved the other dependencies inside the module/Helper folder and fix namespace

What I will do in the next 5 min:
Rename all the Task to Func.

About autoformat:
I'm using productivity power tools, and its do the reorganize imports and autoformat before save. My codestyle now is the original C# prefered. The code in the most files using now a javastyle code format. If you wish I could change my pharenthise options.


tg44:
One more:
Why is the flower check in the Gambling Module? It must be in the games isn't it?

Kwoth:
Leave flowers check for now, I'm still unsure should i use 2 separate currencies for gambling/games or same one.

I will be going back to C# style format, so you shouldn't change your options.

NadekoBot/Modules/Gambling/GamblingModule.cs
```diff
@@ -0,0 +1,71 @@
+﻿using Discord;
+using Discord.Commands;
+using Discord.Modules;
+using NadekoBot.Extensions;
+using System.Linq;
+
+namespace NadekoBot.Modules.Gambling
+{
+    internal class GamblingModule : DiscordModule
```
Kwoth: You have to set this new name in NadekoBot.cs where modules are instantiated. That is why CI is failing.

tg44: Why?
I just renamed a file. Not a class, just a file. The 05c72ff worked correctly.

Kwoth: This class's instance is createdin nadekobot.cs. It's still `new Gambling()` there, needs to be `new GamblingModule()`

Kwoth: Nvm, my bad. It is looking for Gambling.cs. Make sure you've commited your NadekoBot.csproj

Kwoth:
its good, thanks

PR #151, opened 2016-03-25
IMDB search added

tg44:
imdbscraper l222 could be rewriten to async but functionally working


Kwoth:
Thanks a LOT, looks good!

PR #155, opened 2016-03-26
Translator

tg44:
~trans fr>en Ce garçon est fucking bro étonnant
~translangs
~translangs en


LawlyPopz:
1 issue found with the french>english the first letter of each lang has to be caps or it breaks can u fix it to force lower case or convert it to lower and read that :P

Kwoth:
Lawly Popzz demoed the functionality, its great. It works flawlessly. Only thing i will be adding is the typing notification, so you know something is happening while its translating.

PR #156, opened 2016-03-26
Added Beam

LawlyPopz:
Just added some beam stuff!


Kwoth:
weird that you say it doesn't work, looks good to me. I'll accept and fix it

PR #160, opened 2016-03-27
>magicItem

appelemac:
This also removes the pokegame in the Games.cs because it'll get in a conflict with the other pull again if it doesn't :unamused:


DragonOverlord:
yey :weary:

PR #165, opened 2016-03-30
!m repeat

appelemac:
Repeat currently playing song n times (up to 25)


Kwoth:
This doesn't work. It needs to actually repeat the song, not simply add it to the end of the queue.

PR #171, opened 2016-04-01
Fixed pokemon

LawlyPopz:
i have fixed the stats between the pokemon types and added the fairy
type to the types!


Kwoth:
great, thanks a lot

PR #176, opened 2016-04-02
Converter feature

tg44:
Used lib:
http://www.codeproject.com/Articles/23087/Measurement-Unit-Conversion-Library

It is extendable, but I think it's good for a start :)


Kwoth:
thx man, looks good

PR #178, opened 2016-04-02
Moved PokemonTypes to their own json

appelemac:


Kwoth:
all pokemon types need to be moved to json, not just moves. 

You would have a type with multipliers array/object, name, icon, and moves array

tg44:
Please don't take things to the bin folder. Take it to a resources folder, and in the properties window choose the copy option. It's a bad thing that the whole bin folder is not in the gitignore!
(If I will have time, I will do it to the other stuffs too.)

appelemac:
@tg44 idk what Kwoth intends to do with it, so I'll leave it all together for now

Kwoth:
Thx, we'll leave the moving stuff to resources to tg44 if he wants to do it.

Maybe its better to have multipliers as a matrix?

PR #207, opened 2016-04-10
Update PokemonTypes.json

taweko:
Updated Pokemon types to match Bulbapedia's Gen VI type effectiveness chart. (Fixed missing commas).


Kwoth:
Did you test it, does it work?

taweko:
Tested it just now and it works.

Kwoth:
thanks

PR #210, opened 2016-04-11
Cards Against Humanity

DragonOverlord:
-appelemac


Kwoth:
Totaly not untill we get inline whispers (if ever) Its impractical.

Issue #215, opened 2016-04-12
Move excessive info from main github page to wiki


Kwoth:
Make main github page very easy to read, and move instructions and all other stuff to wiki.

readme.md should contain simple questions with simple answers like:

How to get nadekobot?
Go to this server [nadkeologlink], find her in online users and send her your server link

Can i host my own nadekobot?
Yes download updater... or releases or build from source

... etc,

and provide links to comprehensive guides on how to do that stuff on wiki

Kwoth:
done by me 2 days ago

PR #216, opened 2016-04-12
FAQ to readme

DragonOverlord:


Kwoth:
This should be in wiki. See issue #215

PR #230, opened 2016-04-21
Update commandlist.md

Bluscream:


Kwoth:
Commandlist is never updated directly. And these changes don't fit the way other descriptions are written.

Bluscream:
What do you mean with `these changes don't fit the way other descriptions are written` ?

There is no sense or order in the actual commandlist.

Sometimes the usage shows all possible arguments, sometimes not.
The owner only text is only on a part of commands

PR #238, opened 2016-04-26
Custom Reactions

appelemac:


Kwoth:
1. Bring back abalabahaha command.
2. Commands to add/remove/list custom reactions should go to administration module in a separate command file, only thing int he custom reaction module should be the actual custom reactions which are generated from json

PR #240, opened 2016-04-26
Bugfix

tg44:
Currency conversion fixed on ~all locale.


Kwoth:
do we need .mono file?

tg44:
not really but I didn't want to delete/reverse the commit or commit just for remove that file

Kwoth:
thanks

PR #242, opened 2016-04-27
Fix E621

Gaboer:
E621 killed the old code forever, because no user-agent requests apparently is spam. New code works


Kwoth:
Don't use webclient, use httpclient

PR #251, opened 2016-04-30
Better Insults

TheDarkArts:
Added all insults i have running on my bot


Kwoth:
No lol. We have custom reactions in the works.

PR #284, opened 2016-05-08
Update SearchHelper.cs

Iratu:
Fixed danbooru to use a bit more of the build in system to randomly select an image, might work faster if "&limit=1" is added to the image search link but haven't tried this yet.

### DIFF
--- a/NadekoBot/Classes/SearchHelper.cs
+++ b/NadekoBot/Classes/SearchHelper.cs
@@ -222,11 +222,11 @@ namespace NadekoBot.Classes
         {
             var rng = new Random();
 
-            if (tag == "loli") //loli doesn't work for some reason atm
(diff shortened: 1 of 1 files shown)


ShadowNoire:
ive got a further upgrade that replaces sample images with the full size versions

Iratu:
To get full sized images all you need to change is

`var matches = Regex.Matches(webpage, "data-large-file-url=\"(?<id>.*?)\"");`

to

`var matches = Regex.Matches(webpage, "data-file-url=\"(?<id>.*?)\"");`

Kwoth:
won't this reduce the number of allowed tags from 2 to 1?

Iratu:
Hmm, I forgot about that limit, mainly because I upped it to 5 on the booru I run but yes it will reduce it to one, I think it would be nice to add support for the API login so people can increase the search limit to 6 or 12 depending if they have gold or platinum.

PR #311, opened 2016-05-25
Math on rolls

bkohler616:
Preforms calculations within the rolls, as well as support multiple items to roll.
Example: 1d20+3+2d8 would roll a d20, add 3, then add the 2 rolled d8's.
Nadeko will post the evaluated rolls and the equation it will attempt to evaluate so the user can see any issues, then posts the total.


tg44:
These type of commits are bad. Pls config your env to not commit "only white space changed" rows.
Or use: http://stackoverflow.com/questions/3515597/add-only-non-whitespace-changes

bkohler616:
Will fix then. Thanks.

PR #313, opened 2016-05-25
whitelist catblaster from slowmode

Ckath:
please whitelist me from this hell


Ebolo:
https://ebolo.s-ul.eu/3UqaoPIn

appelemac:
Doesn't this just prevent you from being ratelimited

appelemac:
@Kwoth It should be acceptable if we change it to those with Manage Messages perms for the channel

Ckath:
that's the point

Kwoth:
XD

PR #327, opened 2016-06-04
return on error message...

appelemac:


Kwoth:
Oh nice random catch. Thanks

PR #344, opened 2016-06-15
We don't need to download the translate output

darxmorph:
Don't save the translator output to a file.


Kwoth:
Oh wow, that is embarrassing. I blame @tg44 who added translations. :)

Thanks.

tg44:
lol :D I missed that function... nice catch

PR #357, opened 2016-06-21
osu! commands

cbgx:
updated the existing osu! command to accept an optional game mode parameter.

added two new commands for viewing beatmap (song) stats, or a players top 5 scores.
these new commands require an API key, see https://github.com/ppy/osu-api/wiki

an example:
osu! commands sample


Kwoth:
Can you PR dev branch, and remove .sln from PRs?
Otherwise looks ok-ish, and thanks!

I will fix the formatting a bit after accepting.

PR #365, opened 2016-06-26
Fixed ~rule34

Soundofdarkness:
Fixed the issue with ~rule34 displaying .webm files. Bot is now displaying a message if it gets a .webm.
I did not add an automatic new search, because it will result in a permanent search and probably crash while used with tags that are only including .webm files.


Soundofdarkness:
Wrong Branch .

PR #366, opened 2016-06-26
Fixed ~rule34

Soundofdarkness:
Fixed the issue with ~rule34 displaying .webm files. Bot is now displaying a message if it gets a .webm.
I did not add an automatic new search, because it will result in a permanent search and probably crash while used with tags that are only including .webm files.
(Redid pull request because the first one was in the wrong branch)


appelemac:
What if someone wants to see that webm?

Soundofdarkness:
I could add an config option for it . But at least from the the users from my bot were annoyed by the webm

Kwoth:
nah, not needed

PR #405, opened 2016-07-14
Patch 2

miraai:
Added few steps on how to set up ffmpeg on Jessie, since it is not available on Debian 8 Jessie, but in Backports.


Kwoth:
Ty

PR #414, opened 2016-07-17
Update LinuxSetup.md

miraai:
- Added fix for Linux Mono installation where the bot won't run or it would just show random wall of string.
- Added mono fix for CentOS 7 and Fedora 19.


Kwoth:
thanks

PR #422, opened 2016-07-19
Sped up .prune drastically

Loolzy:
.prune has been sped up by using DeleteMessages method instead of a foreach-loop and deleting them one by one


Kwoth:
I'm working on it myself atm, thanks for finding the method though :D

PR #449, opened 2016-07-25
Fix destruction of text channels being counted as creation of text channels

Tibo442:


Kwoth:
It always seemed there is a lot of voice channels.

Thanks a lot.

PR #464, opened 2016-07-28
Basic audio cache

Nitix:
I have added a basic audio cache from SoundCloud and YouTube, it can be deactivated with config file and have a specified max audio cache size.

When the cache has take all space available, it just stop caching new file, that's why it's a basic audio cache. He didn't check the last time it was played or how many times it was played.

Cached file are encoded with Opus codec and use video id from youtube/soundcloud for the filename and are stored in a user specified folder.

The cache system add new attributes in SongInfo that have default value for non Youtube/SoudCloud files, and they are not handled by the cache system (It would be stupid to cache and file that is already in the server). 

It would be pretty easy to add new provider to the cache system.

I have tested the audio cache in Windows and Ubuntu.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,11 +7,15 @@
 *.userosscache
 *.sln.docstates
 *.pfx
+*.tmp
+*.TMP
+
 obj/
 # User-specific files (MonoDevelop/Xamarin Studio)
 *.userprefs
 **/Bin/Release/
 **/Bin/PRIVATE/
+**/Bin/Debug/cache/
 !**/Bin/Debug/opus.dll
 !**/Bin/Debug/libsodium.dll
 !**/Bin/Debug/Nito.AsyncEx.Enlightenment.dll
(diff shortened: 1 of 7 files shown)


appelemac:
:ok_hand: :+1:

Nitix:
Don't merge it now, I found a possible flaw, if two servers or more play the same music at the same time, ffmpeg will try to write in the same file at the same time. I will add a temp unique id and check if cache file is already here.

Kwoth:
Why not simply use the current timestamp?

Nitix:
Ok, I changed to timestamp. If you have others suggestions before merging,  feel free to ask.

Kwoth:
Is this not possible anymore due to how changes to how music is saved?

Nitix:
I can adapt it to the new format easily, It just because i moved the buffer function to a specific class, but I can still do this.

PR #479, opened 2016-07-29
Add transaction for database, FIX #471

### LINKED ISSUE #471: SQLite speed needs to be drastically increased
Every DbHandler call will create a a new connection to the database, which increases the duration of each insert by 100%. (around 200q/s vs 100q/s)

Since every insert statement starts a transaction and commit it right after, this will drastically reduce the number of queries per second. If I could manage to make a queue with queries needed for execution every-so-often, the queries can go from 200/s up to 65k/s

https://www.sqlite.org/faq.html#q19

I will probably drop Sqlite-net for something else

Nitix:


Kwoth:
got it on a branch with the same name as yours

Issue #470, opened 2016-07-28
!!ap should use a random song out of the first 3-5 related ones in order to prvent looping


Kwoth:
Sometimes autoplay can get stuck in a loop where song A is the first suggestion for song B and song B is first suggestion for song A, so the bot gets stuck going back and forth between those 2.

PR #482, opened 2016-07-31
Prevent looping of autoplay by using random video from first 5 related.

Tibo442:
Adds #470


Kwoth:
Omg, really? I literally just finished this. If you were just 5 minutes quicker, while i was checking which issue to solve.

Thanks a lot for trying to help though :\

PR #483, opened 2016-07-31
Make autoplay tracks be queued from bot and not someone else

Tibo442:
And let's not make the same changes in the same minute :P


Kwoth:
there is Server.CurrentUser.

Tibo442:
Oh, didn't know about that, reverted commit and re-commited with queuer.ServerCurrentUser

Kwoth:
thanks

PR #484, opened 2016-07-31
Change how we handle audio file buffer, FIX #478

### LINKED ISSUE #478: Radio or long music can fill your disk entirely
In dev branch, the way on how is handled buffer can fill up the disk. When buffer is end to client it's not removed from file, so radio that are unlimited will create a really big file, as data are saved in raw byte in the file.

Nitix:
Now it would create small files 20Mb and rotate them by using an identifier (type long).

When a file has been read, Song will be delete it.

### DIFF
--- a/NadekoBot/Modules/Music/Classes/Song.cs
+++ b/NadekoBot/Modules/Music/Classes/Song.cs
@@ -32,7 +32,6 @@ namespace NadekoBot.Modules.Music.Classes
         public SongInfo SongInfo { get; }
         public string QueuerName { get; set; }
 
-        private bool bufferingCompleted { get; set; } = false;
         public MusicPlayer MusicPlayer { get; set; }
 
         public string PrettyCurrentTime()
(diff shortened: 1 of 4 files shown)


Kwoth:
I'm testing it. Does it pause downloading when it hits 200MB ahead of what's been played like it does right now?

Nitix:
I put the 100 MB limit back and fixed the incorrect deletion.

Kwoth:
First time i ran it, music thread crashed with the message "File can't be found x12312312312-0".

Afterwards it worked flawlessly

NadekoBot/Modules/Music/Classes/Song.cs
```diff
@@ -73,78 +72,22 @@ public Song SetMusicPlayer(MusicPlayer mp)
...
         internal async Task Play(IAudioClient voiceClient, CancellationToken cancelToken)
         {
             var filename = Path.Combine(MusicModule.MusicDataPath, DateTime.Now.UnixTimestamp().ToString());
 
-            var bufferTask = BufferSong(filename, cancelToken).ConfigureAwait(false);
+            SongBuffer sb = new SongBuffer(filename, SongInfo, skipTo);
+            var bufferTask = sb.BufferSong(cancelToken).ConfigureAwait(false);
 
-            var inStream = new FileStream(filename, FileMode.OpenOrCreate, FileAccess.Read, FileShare.Write);
+            var inStream = new FileStream(sb.GetNextFile(), FileMode.OpenOrCreate, FileAccess.Read, FileShare.Write); ;
```
Nitix: This shouldn't create an error, even if SongBuffer hasn't create the file yet.

Kwoth:
From what i tested, it works. Thanks

PR #485, opened 2016-08-01
Minimal last.fm module

moist-anus:
Added a basic last.fm module to display scrobble information.


blitz4694:
probably should take out nadekobot.sln

Nitix:
Be careful with space and tabulation, it can break indentation (you can see it on github).
Also you should move your LastFm classes from Nadeko/Classes to it's submodule.

moist-anus:
can do, thanks

PR #498, opened 2016-08-02
Added Ability to reset flowers from a given ID

Soundofdarkness:
Ability to remove flowers from a given User, who is not on the server anymore. 
Usage : `$reset [ID]`


Kwoth:
just use $take

PR #502, opened 2016-08-02
Added ID Usage for $take

Soundofdarkness:
$take is able to accept an ID now.


Soundofdarkness:
*is able to accept an ID now

PR #534, opened 2016-08-05
fix cbl and cubl commands

MasterDomino:
added verification on cubl soo if key exist remove it if doesnt inform client.

ps. ppl told me that you dont like cleaning code soo tell me if you want me to pr again :P


Kwoth:
111 files changed?

PR only with actually changed files.

PR #545, opened 2016-08-06
add error coloring and fix some redundant vars

MasterDomino:
added memory monitoring


Kwoth:
.ResetColor() doesn't help a lot as it takes same amount of space :D

Maybe you could write an extension method .WriteInColor(string,ConsoleColor); which would write that text in color and then reset the color

MasterDomino:
its not meant to avoid space usage its meant to cover general C# conventions but if you wish :P

PR #547, opened 2016-08-06
add error coloring and fix some redundant vars

MasterDomino:
added memory monitoring
same as #545  but added WriteInColor


Kwoth:
- Too many changed files (Don't change everything to use new method, only the things where I used it in the code already are enough)
- Keep 1 feature per PR (add memory in a separate PR, but i don't think we really need it)

PR #556, opened 2016-08-08
The anti-triggered edit

GitBranch:
Added in zh-TW before someone detonates 3000 more nuclear bombs on everything

(I have no idea if this actually works or not since I don't have a DE set up)


blitz4694:
dont pr into master
do it into dev

GitBranch:
OK

PR #557, opened 2016-08-08
The Anti-triggered Edit™ 2: Electric Boogaloo

GitBranch:
I pulled into the wrong branch. Now this should be correct.

I'm not very familliar with git yet so I don't know how to change branches.


Kwoth:
1. Are you sure this is correct? Zh-TW sounds like taiwan, not traditional version
2. chinese should stay "zh-CN"

MasterDomino:
1. https://en.wikipedia.org/wiki/Zh-TW <=trad. chinese
2. https://en.wikipedia.org/wiki/Simplified_Chinese_characters <= simplified one

ps.Sorry for linking

GitBranch:
1. Yes, it is correct, zh-TW is indeed Traditional Chinese of the Taiwanese variety. zh-HK is Chinese (Hong Kong) but it's almost identical, and also missing from Google translate.

2. It's best not to simply put "chinese" as zh-CN just to remain neutral, even though Google translate does that. However if you don't think so I'll keep chinese-simplified and add back chinese as zh-CN.

Kwoth:
For was majority of users, it is a convenience to have chinese-simplified as chinese. Just remove "-simplified", or add another one as "chinese" with zh-CN too.

GitBranch:
I guess it is technically the majority (1.3 billion > 40-50 million)... I'll add it back

Kwoth:
ty

PR #568, opened 2016-08-11
Invariant Culture for parsing in osu commands

Seregy:
Parsing some values from osu response would fail if local culture used comma as decimal separator. Invariant Culture fixes it


Kwoth:
nice, thanks

PR #597, opened 2016-08-27
Updating unclear instructions with tmux

Zeptaxis:
I had to google because I thought the Ctrl+B+D thing wasnt' working, so I propose this small update to make it more clear that you need to press Ctrl+B and then D and not all at the same time.


appelemac:
If you're doing this, please fix the typo of detach in the next line too ^.^

Kwoth:
Thanks, In the future, PR into dev.

PR #606, opened 2016-08-30
Unitconversion

appelemac:


appelemac:
btw, this isn't in DB because A: want to make some other changes, B. want to modify a minimal amount of files.
Don't merge yet

PR #629, opened 2016-09-11
Update to OSX Guide.md

fearnlj01:
Should be the same functional edit as what Xnaas suggested for the linux guide, same spelling mistake and keyboard combo.


miraai:
Added. Thanks.

PR #649, opened 2016-09-21
Add shards support

Nitix:
Having a DBMS will be a must use, as guild can move between shards. SQLite will not be able to share data between shards.

### DIFF
--- a/src/NadekoBot/NadekoBot.cs
+++ b/src/NadekoBot/NadekoBot.cs
@@ -42,16 +42,20 @@ namespace NadekoBot
 
             _log.Info("Starting NadekoBot v" + typeof(NadekoBot).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion);
 
+            //Setting up credentials
+            Credentials = new BotCredentials();
+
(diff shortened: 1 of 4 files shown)


Kwoth:
We need a bit more sophisticated system than this.

I wanted a ShardedClient to create a connection for each shard, and have same methods normal discord client has. For example shardedClient.GetGuilds() would do shards.SelectMany(shard=>shard.GetGuilds()) - shard being DiscordClient.

Credentials could contains a number of shards, or we could actually just automatically spin up a new shard for every 2500 servers (i think that was the limit). Creds could also starting shard id.

For example, you want to run 4 shards per physical server, you would setup 1 bot, set its max shards to 4 and starting shard id to 0. That's first 10k guilds. On another server, you set 4 shards again and starting shard id to 4

Nitix:
You can't add shards dynamically You have to set the total number of shards at the first shard. And the benefits of shards is to distribute the bot across multiples servers. I'm not sure adding complexity will help, just running multiple instances of Nadeko should be a better solution. 

Adding shardedClient will have to change all call to APIs

Kwoth:
Closing as I've added a different solution

PR #665, opened 2016-09-30
Change ulong to long to be compliant with dbms

Nitix:
PostgreSQL is suppported, SQLServer should also work (not tested)


Kwoth:
Looks good, although its dangerous to merge this today. I'll merge it tomorrow/in 2 days.

Btw any reason you used Convert instead of just (long)someulongnumber

Nitix:
I don't know, even more casting would even be more appropriate as it's exception free.
Anyway I need to resolve conflict

PR #667, opened 2016-10-02
 Fix Foreign key with delete on cascade

Nitix:
Don't ask why, but it's working, I made almost no changes.

SQLite is very shitty, it doesn't like migrations :(


Nitix:
wait, there is something wrong, gonna check

PR #670, opened 2016-10-02
Customreactions

appelemac:
Left on the TODO: the custom formats need to be implemented in a way we agree on


appelemac:
:unamused: I'm only gonna prepare this for merge once

PR #672, opened 2016-10-04
Made a types command by just copying and pasting

blitz4694:


Kwoth:
we probably won't have any more releases of 0.9, so this would be pointless. Isn't there "typelist" command already btw

PR #674, opened 2016-10-05
0.9 to 1.0 migration

Nitix:


src/NadekoBot/Modules/Administration/Commands/Migration.cs
```diff
@@ -0,0 +1,220 @@
...
+                    blacklist.AddRange(oldData.ChannelBlacklist.Select(channel => new BlacklistItem() {ItemId = channel, Type = BlacklistItem.BlacklistType.Channel}));
+                    blacklist.AddRange(oldData.UserBlacklist.Select(user => new BlacklistItem() {ItemId = user, Type = BlacklistItem.BlacklistType.User}));
+                    botConfig.Blacklist = new HashSet<BlacklistItem>(blacklist);
+
+                    //Eightball
+                    botConfig.EightBallResponses = oldData._8BallResponses.Select(response => new EightBallResponse() {Text = response}).ToList();
+
+                    //NOW save it
+                    botConfig.MigrationVersion = 1;
+                    uow.CompleteAsync();
```
Kwoth: await is missing

src/NadekoBot/Modules/Administration/Commands/Migration.cs
```diff
@@ -0,0 +1,220 @@
...
+                }
+            }
+
+            private class MigrationException : Exception
+            {
+            }
+
+            protected class CommandPrefixes0_9
```
Kwoth: this should be in a separate files/namespaces

src/NadekoBot/Modules/Administration/Commands/Migration.cs
Nitix: No, method is not async, command is but not this method

PR #696, opened 2016-10-16
fix infinite getter recursion

jvitkauskas:
Fix a typo


Kwoth:
wow. Thanks. I pushed a fix in 1.0 branch

PR #708, opened 2016-10-21
Guides ->  1.0

fearnlj01:
It took so long after the last mention since I was too stubborn to just do this from a local git client for too long.

Linux and OS X guides both have things built one by one for now and I'll re-release the windows updater once I get the A-OK for not having to build things one by one.


Kwoth:
ty

PR #720, opened 2016-10-24
Fixed Unhandled Exceptions Crashing NadekoBot.

AraHaan:
Whenever a Exception happens while trying to Login NadekoBot Crashes. With this it prevents a crash and tells you the exception message.

### DIFF
--- a/src/NadekoBot/Program.cs
+++ b/src/NadekoBot/Program.cs
@@ -9,7 +9,7 @@
             }
             catch (System.Exception e)
             {
-                System.Console.WriteLine("NadekoBot Crashed with Exception: {0}", e.Message);
+                System.Console.WriteLine("NadekoBot Crashed with Exception: {0}", e);
             }
         }
     }
--- a/src/NadekoBot/Program.cs
+++ b/src/NadekoBot/Program.cs
@@ -2,7 +2,15 @@
 {
     public class Program
     {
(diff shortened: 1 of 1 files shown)


Nitix:
Not sure but having the the name of the exception and the stack trace in the console, may help to troubleshoot the crash.

AraHaan:
hmm true.

Kwoth:
it already prints errors in the console, so no need for this, also, don't use console.writeline but Logging

AraHaan:
Actually even then your exception handling is not handled at the time these exceptions are thrown so then mine catches it otherwise Nadekobot will actually crash with them unhandled because of a few core issues that can happen so, this wors, I could use your logger though. But yeah any Exception it has while and before it can login (which can happen) can actually come unhandled because of this:

    internal Task LoginAsync(TokenType tokenType, string token) =>
        Task.WhenAll(Clients.Select(async c => { await c.LoginAsync(tokenType, token); _log.Info($"Shard #{c.ShardId} logged in."); }));

can actually fail. As well as anything Else like if discord.net fails to create a websocket that it would Crash Nadeko as well. So making it not crash unhandled is a must.

AraHaan:
Well Edited Wrong file now that I look in NadekoBot.cs that is the file I should have edited for this.

PR #723, opened 2016-10-24
Fix for readthedocs and Kwoth's Magic ✨

samdivaio:
United all commands for readthedocs


samdivaio:
No need that, readthedocs are just lazy at updating things

PR #741, opened 2016-11-06
Playlist cancel

appelemac:


Kwoth:
Did you test this? Please test it heavily, try with all kinds of playlists - scpl, lopl, pl, try queuing normal songs afterwards, try canceling multiple playlists in a row. When i wake up I'll merge this

PR #753, opened 2016-11-08
dots gone

miraai:


Kwoth:
you missed the shards one below

PR #758, opened 2016-11-11
Various changes ¯\_(ツ)_/¯

fearnlj01:
Summary of Changes:
- Betroll won't give you the greater than 90 bonus for rolling 90
anymore
- Send usage updated for clarity
- I missed some stuff when the  `config.json` stopped having "null"
everywhere, this fixes it, *I think.*
- more whitespacing in the linux guide for people like Kwoth who want to
just be able to copy paste commands

I'm kinda just guessing this'll work so if it doesn't then, "woopsie"


Kwoth:
👌

PR #766, opened 2016-11-14
Add Yande.re search

fkndean:


Kwoth:
Hey, thanks for another pr!

If they do have an api, since i see you're using .xml, maybe use it normally instead of regex. and if you could check if there is 2 tag limit on this one too for sure. If tag limit is greater than 2, then use one for Order:random, and get only first post - much less expensive. 1000 is too much. Also add it to ~hentaibomb.

fkndean:
Doing it now and looking further into it, I found that it's pretty much the same as danbooru. I don't use these sites lol. Going to be pushing commits soon.

fkndean:
image

fkndean:
Removed safeyandere due to request.

Kwoth:
ty 👌

PR #771, opened 2016-11-18
some fixes and cleanup

samdivaio:
Removed the extra `Name#Discrim: #` in command `.uinfo`
Removed the bold text for channel topic to use the default channel topic style,
Changed bolt to italic for server custom emojis
Removed the bold texts for server icon url to make it show the server image at the end and to make it more helpful.
Added emoji to various help commands to make it look better.
Also added space between Nickname and ID if the user has nickname set and uses `.uinfo` command.


samdivaio:
fixed emoji only worked for PC discord clients and showed random texts on iOS and Android clients

PR #779, opened 2016-11-20
Initial Imgur API commit

fkndean:
The initial start of a great beginning lel.


fkndean:
Created Imgur in its own class file.
Added `~rgi`

>Looks for the top and viral images which takes arguments page number, followed by a date range (day | week | month | year | all)

PR #806, opened 2016-11-27
Magic The Gathering API

fkndean:
Since Hearthstone, why not? :) Doesn't need a API key.

### DIFF
--- a/src/NadekoBot/Modules/Searches/Searches.cs
+++ b/src/NadekoBot/Modules/Searches/Searches.cs
@@ -196,6 +196,20 @@ $@"🌍 **Weather for** 【{obj["target"]}】
 
         [NadekoCommand, Usage, Description, Aliases]
         [RequireContext(ContextType.Guild)]
+        public async Task Google(IUserMessage umsg, [Remainder] string terms = null)
+        {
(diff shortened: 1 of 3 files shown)


fkndean:
image

This was before the message output was removed in `d363af4`

src/NadekoBot/Modules/Searches/Searches.cs
```diff
@@ -210,6 +210,51 @@ await channel.SendMessageAsync($"https://google.com/search?q={ WebUtility.UrlEnc
...
+            string response = "";
+            using (var http = new HttpClient())
+            {
+                http.DefaultRequestHeaders.Clear();
+                response = await http.GetStringAsync($"https://api.deckbrew.com/mtg/cards?name={Uri.EscapeUriString(arg)}")
+                                        .ConfigureAwait(false);
+                try
+                {
+                    var items = JArray.Parse(response).Shuffle().ToList();
+                    var images = new List<Image>();
```
Kwoth: why are you adding them to list.

Maybe you can just print the stats like you did before and put image link in there too. It will auto-embed. I think we can do much better stuff with rich embeds when we get them though

fkndean:
Done @Kwoth hope you approves.

PR #807, opened 2016-11-27
Add formatting output with scr as well as the actual output reaction.

fkndean:
So it's easier to just copy and paste it with the formatting for the show custom reactions.


Kwoth:
isn't there Format.Escape or something like that, what you did wouldn't look so pretty

fkndean:
image

Kwoth:
`Discord.Format.Sanitize` <- that's what i meant

PR #808, opened 2016-11-28
Adapt the PokeGame from 0.9 to 1.0

Taknok:


Kwoth:
actually looks legit, ill take a look later today/tomorrow

Taknok:
Cool, thank you :)

PR #819, opened 2016-11-29
Help output as embed

fkndean:
Make everything look nice!
image


Kwoth:
why are there so many icons xD

Remove author, remove footer, remove icon, remove thumbnail, put the description to description, command names to Command field which is inline with usage field

fkndean:
options. options. options. x)
Doing some edits now and doing error output as well.

fkndean:
This job has been picked up by @samdivaio

PR #841, opened 2016-12-02
Guides update

samdivaio:
- branch changed from 1.0 to dev (Contribution Guide)
- FAQ improved/ updated and linux certs question for old mono issue is now removed.
- more info added to DB guide (its never enough)
- few changes to upgrading guide like links to latest guides and lil details.


miraai:
You are the best, bo <3

PR #847, opened 2016-12-04
Update CommandHandler.cs

fkndean:
Make errors output in channel. ily kwoth

image


Kwoth:
no need for server and channel, also this can't be added right now because there are cases where we have really long errors (with full stack trace)

PR #850, opened 2016-12-04
Update LogCommand.cs

fkndean:
Lets actually make the log have a proper date format.


samdivaio:
logserver uses default server time so I don't think it is necessary to add the whole long data and time format and with UTC

Kwoth:
just change it from .Now to .UtcNow

fkndean:
Honestly personal preference 📦

samdivaio:
If its possible to include choosing time zone/server(guild) that would be useful even with date option to choose from that would be best.

PR #854, opened 2016-12-05
Update Utility.cs .whpl command

fkndean:
image


Kwoth:
I don't like that format. Maybe just a single field name "Users who are playing XXXXXX" with users in the value, and no thumbnail.

Kwoth:
I don't like this

PR #858, opened 2016-12-05
Initial Multilang Support

fkndean:
This is the beginning.
Work in progress.


src/NadekoBot/Modules/Administration/Commands/Migration.cs
```diff
@@ -102,6 +102,7 @@ private void MigrateDb0_9(IUnitOfWork uow)
                     var greet = (long)reader["Greet"] == 1;
                     var greetDM = (long)reader["GreetPM"] == 1;
                     var greetChannel = (ulong)(long)reader["GreetChannelId"];
+                    var language = (string)reader["Language"];
```
Kwoth: wtf? This shouldn't be here

fkndean: oh oops

PR #859, opened 2016-12-06
Update Pokemon.cs

fkndean:
image


Kwoth:
This is not how it should be implemented, ill add it

PR #861, opened 2016-12-07
fix for everyone mention

samdivaio:
sometime logserver (discord,net) makes the (roles added or removed) single role name to list of all the roles including everyone mention triggering the mention to everyone. hope this fixes that issue until the bug is fixed


Kwoth:
fikxall bugs sam

samdivaio:
Isn't that discord.net bug?

PR #909, opened 2016-12-20
fixed ~konachan results

Zylon:
Konachan results didn't have "Http:" in front. so they would return as:
//konachan.com/image/...


Kwoth:
thx but already fixed in the upcoming branch

PR #930, opened 2016-12-28
Update Searches.cs to support gif avatars

sansfont:
Gif avatars are now a feature of discord with discord nitro. 

All gif avatars' id start with "a_"

Check if the first characters of the string are a_ and if so, the url will be https://images.discordapp.net/avatars/[user id]/[avatar id].gif

If this is not the case, use the old method.


Kwoth:
this doesn't compile, but thanks for the heads up

PR #969, opened 2017-01-12
Updated regex for playlist, cleaned up regex fix earlier

shikhir-arora:
- Regex from earlier is slightly cleaner. If you are using it to queue music just with `!!q`, you can use any YouTube keyword, any standard http/https YouTube URL, YouTube short URL's that contain a string of letters or `_` or `-` which wasn't the case in the earlier Google/YT API regex. And many of those short codes are still just strings of letters unlike the full URL's. So you can use: keywords, https://youtu.be/dQw4w9WgXcQ, or https://www.youtube.com/watch?v=dQw4w9WgXcQ (http or https) however & or other operators will show as wrong URL after the ID. That ensures that it is, again, only picking up the ID.

- Because of this one can also queue __seamlessly__ using the video ID:  `!!q dQw4w9WgXcQ` 

- Clean regex added for playlist. This is a _complete_ regex that will handle malformed playlist URL's, as well as most expressions. Because of this, using `!!playlist` (or `!!pl`) with a correctly formed playlist URL such as: https://www.youtube.com/playlist?list=PL4lEESSgxM_5O81EvKCmBIm_JT5Q7JeaI

will result in some verbose outputs every time the Google API cannot pull the track. In this way it is kind of good as it shows you an output every time it cannot get the playlist URL for any reason. **It is common, sort of out of control with massive queued playlists** but I tested it and only 20 or so weren't found out of close to 1,000.

- Again, you can very easily and cleanly (almost preferred) use the playlist ID (from above): `!!pl PL4lEESSgxM_5O81EvKCmBIm_JT5Q7JeaI` will work great

- You can of course use keywords `!!pl keywords` as usual; works just fine. Again any URL's in large queue's it cannot get it will output as verbose output (will ***not***) kill the bot session in any way. But it will let you know. 

----------

**TESTS DONE**

- [x] Queued multiple songs from SoundCloud and YouTube

- [x] Large playlist from YouTube (to the point of ratelimit!) - verified skipping `!!n` with a large amount, >300 -- this causes a bit of a delay but it is a very extreme example of using `!!n` to skip hundreds of songs in. Nadeko doesn't spam excessively either and deletes messages as expected.

- [x]  Verified moving songs in large queues (different positions), mixing a loaded playlist with a new playlist, removing and skipping. Also mixing in. 

- [x] Cleared queue using `!!stop` and immediately after reloaded - first with playlists, then repeating this while checking commands and regular music. Verified `!!destroy` (and subsequent re-queue) without restarting the bot. Finally, concluded with resuming `!!radio` stream, which I had going earlier for 8-12 hours on this codebase with no issues.  😄 

--------

(this does ***NOT** fix `!!ap` --> that issue was before any of this. I made a few changes locally but have to do more troubleshooting)

### DIFF
--- a/src/NadekoBot/Services/Impl/GoogleApiService.cs
+++ b/src/NadekoBot/Services/Impl/GoogleApiService.cs
@@ -51,7 +51,7 @@ namespace NadekoBot.Services.Impl
             return (await query.ExecuteAsync()).Items.Select(i => i.Id.PlaylistId);
         }
 
-        private readonly Regex YtVideoIdRegex = new Regex(@"(?:youtube\.com\/\S*(?:(?:\/e(?:mbed))?\/|watch\?(?:\S*?&?v\=))|youtu\.be\/)([a-zA-Z0-9_-]{6,11})", RegexOptions.Compiled);
+        private readonly Regex YtVideoIdRegex = new Regex(@"(?:https?://)?(?:www\.)?(?:youtube\.com/watch\?v=([^&\n]+)|youtu\.be/([a-zA-Z\d_-]+))", RegexOptions.Compiled);
 
         public async Task<IEnumerable<string>> GetRelatedVideosAsync(string id, int count = 1)
         {
--- a/src/NadekoBot/Services/Impl/GoogleApiService.cs
+++ b/src/NadekoBot/Services/Impl/GoogleApiService.cs
@@ -38,7 +38,9 @@ namespace NadekoBot.Services.Impl
             if (count <= 0)
                 throw new ArgumentOutOfRangeException(nameof(count));
 
-            var match = new Regex("(?:youtu\\.be\\/|list=)(?<id>[\\da-zA-Z\\-_]*)").Match(keywords);
+// a bit extra here so it accepts malformed playlist URL's. may throw error because it's expecting the contents after the ?list here, but we are allowing a full regex. although you get some boilerplate warnings, everything loads very quickly and properly.
+            
+            var match = new Regex(@"^(?:https?:\/\/)?(?:www\.)?(?:youtu\.be\/|youtube\.com\/(?:embed\/|v\/|watch\?v=|watch\?.+&v=))((?:\w|-){11})(?:&list=(\S+))?$").Match(keywords);
             if (match.Length > 1)
             {
                 return new[] { match.Groups["id"].Value.ToString() };
@@ -50,6 +52,7 @@ namespace NadekoBot.Services.Impl
 
             return (await query.ExecuteAsync()).Items.Select(i => i.Id.PlaylistId);
         }
+//proper regex that returns correct video IDs for normal youtube links every time
 
         private readonly Regex YtVideoIdRegex = new Regex(@"(?:https?://)?(?:www\.)?(?:youtube\.com/watch\?v=([^&\n]+)|youtu\.be/([a-zA-Z\d_-]+))", RegexOptions.Compiled);
 


src/NadekoBot/Services/Impl/GoogleApiService.cs
```diff
@@ -38,7 +38,9 @@ public GoogleApiService()
             if (count <= 0)
                 throw new ArgumentOutOfRangeException(nameof(count));
 
-            var match = new Regex("(?:youtu\\.be\\/|list=)(?<id>[\\da-zA-Z\\-_]*)").Match(keywords);
+// a bit extra here so it accepts malformed playlist URL's. may throw error because it's expecting the contents after the ?list here, but we are allowing a full regex. although you get some boilerplate warnings, everything loads very quickly and properly.
+            
+            var match = new Regex(@"^(?:https?:\/\/)?(?:www\.)?(?:youtu\.be\/|youtube\.com\/(?:embed\/|v\/|watch\?v=|watch\?.+&v=))((?:\w|-){11})(?:&list=(\S+))?$").Match(keywords);
```
Kwoth: shoudn't `(?:&list=(\S+))?` be `(?:&list=(?<id>\S+))?`

Like this it will never match a group with "id". That might be the error you're getting. Also i don't like the idea of having errors in the console. Users get very upset. Also `^(?:https?:\/\/)?(?:www\.)?` seems unnecessary. We don't care what's preceding the url. it could be `oajn=32h-=39ubf92-3fb329-youtube.com?list=123123123` and it should still match it. No need to force people to not make copy-paste/ typing errors.

Kwoth:
Please post a screenshot of the error in the console you're getting with playlists using this code. 

Also, I don't understand what you mean by 
`So you can use: keywords, https://youtu.be/dQw4w9WgXcQ, or https://www.youtube.com/watch?v=dQw4w9WgXcQ (http or https) however & or other operators will show as wrong URL after the ID.`

We must be able to let users use links with &, &t , &list=, etc after the video id. It's a vital part of user experience (not having to manually edit youtube links they copy-paste)

shikhir-arora:
I understand where you are coming from; feel free to modify it as you stated in your prior comment. If I queue a large playlist using a keyword, say 250 songs, because of the way YouTube's API works it will almost never return 250 songs perfectly. This has always been the case. However, it is usually not many (2 or 3) that do not end up going in the queue. YouTube API records the error in your Google account. 

All the regex did for playlists is output (as it's loading a queue) a single sentence in the console (say, forbidden for some license with a 403 error or 401). These stay within tmux and the bot continues to queue. So if I looked up a keyword for `!!playlist house music` which got a 250 song list with 5 that could not be queued, there would be 5 of these printed in the console. Again, you can still `!!lq` and the queue is building. So I only said it was good from a developer point of view because you can see how many are not loading and it stays within the tmux session. When the bot is done with the queue, the command completes, updates, etc. as usual. 

This is only for playlists. I did simplify the regex for the regular `!!q` with this, and it will accept what I stated. It won't affect playlists; just regular tracks. If you put an `&` or t?=192 (some timestamp) next to this yes it will say "cannot find song" for the single queue. This ensures that when you use `!!q keywords` the URL is parsed without extraneous characters, which caused errors before because the bot wouldn't be able to parse it. However, this is not any different from the last commit that went out re: single URL queues. I just cleaned it up so that you can easily use (for both playlist and regular URL's) the ID *directly* which works seamlessly. 

[I did not have to shut down the bot during any of my tests for the last few days in any commits, including radio playlists) -- however, I understand the concern with users seeing errors if they put an & symbol or whatnot.

shikhir-arora:
Here is a screenshot. Note this is what happens before with large playlists; in this case I looked up a playlist with `!!pl keyword` which was 500 songs and it queued 464. Note that `!!lq` and the first song begin playing and outputting quickly; that time is just for the entire queue. Basically, it's going through each URL in the queue and ensuring it won't cause an error __later__ such as suddenly stopping when it gets there, skipping, or just playing without sound. Each URL is checked; if there's an error on the backend YouTube API it is shown beforehand. The returned URL's are all essentially guaranteed to work. This was the same idea behind the single queue from yesterday, but to a much smaller scale: these errors happened before; but I see where the user experience comes to play vs. developer.

(There are not errors again that affect Nadeko, just backend errors from YouTube API so your bot works fine for everything)

screen shot 2017-01-12 at 7 32 08 pm

shikhir-arora:
TODO JAN 13

- [ ]  Cleanup verbose commands for regex and handle extraneous characters from copy/paste etc.

Kwoth:
No need for this since regexes seem to work flawlessly atm

PR #972, opened 2017-01-13
Updated downloaders to close automatically, so auto restart doesnt need confirm.

theoddball:
Title. When set to auto update, it makes you press a key in console to update, which prevents auto restart


Kwoth:
Thanks, PR master please

PR #1007, opened 2017-01-26
Add autoinstallers to guide

theoddball:


Kwoth:
We need @samdivaio probably to properly add this to the guide, this being the main way for people to install, and then the long guide if it didn't work

theoddball:
Okiedokie. Working on more automation of it right now so this info is out of date.

PR #1029, opened 2017-01-31
Youtube Reopen locally

DaYWhiTe:
Record the songs we've opened on YouTube then open it locally when you reopen


DaYWhiTe:
Even as an option. 

Can be opened and closed with these commands;

!!youtubesave on/off | !!ys on/off Are you recording songs from YouTube?

PR #1084, opened 2017-02-23
Fix specials characters for memegen

Nitix:
Using specials characters like ? or # cause the bot to post incorrect link. The link shouldn't URI encoded by the bot as discord do this on their end.

I wrote the replace function using https://github.com/jacebrowning/memegen as reference for specials characters.


Kwoth:
Thanks a lot dude <3

PR #1098, opened 2017-03-02
Weather and YouTube Playlist queue changes.

samdivaio:
- Playlist queue emojis added again.
- Weather changes includes minor cosmetic changes including changing `km/h` to `m/s` and adding text `utc` to the sunrise and set times, weather api icon in the footer as a openweatherapi thing, texts field looks changed to bold for mobile clients.


samdivaio:
Clickin on city name can now redirect to website with more detailed openweather info.

Kwoth:
Don't bold the field names. It's just too much. It's a bug on mobile devices, just wait until discord fixes it.

samdivaio:
I think pc needs to be fixed not mobile devices also if they remove bold to field name we will remove it then, currently it messes up field names on some high density display of android if there is a emoji in field name. :/

PR #1100, opened 2017-03-02
Added support in .qsearch output, support for all cases, add ContainsNoCase Extension  

shikhir-arora:
~- Added and tested support for embeds / embed content search in `.qsearch` output so the result isn't plaintext JSON as it is currently when the quote is an embed type~

- Minor change with [the keyword - displays as lowercase as it is nicer to read than the upper invariant default pulled from the database](https://github.com/shikhir-arora/NadekoBot/blob/c7e694fc1d01fe07d9e1b0fd92b769ef1ff81678/src/NadekoBot/Modules/Utility/Commands/QuoteCommands.cs#L93)  

~- Added and tested so results for the text (content of the quote) are case-insensitive freely for all parts (i.e. "That was a GREAT comment" as a quote saved under the keyword "test" will work with searching `.qsearch test great` or `.qsearch test GReAt` or `.qsearch test GRE` or `.qsearch test grEAt` etc. as it should be because people don't remember the case in their quote content; obviously the keyword must match exactly as always)~

- Added and implemented `ContainsNoCase` extension of the `String.IndexOf Method (String,StringComparison)` method for fast comparison through `StringComparison` types CurrentCultureIgnoreCase, 	InvariantCultureIgnoreCase, OrdinalIgnoreCase

- Implemented `ContainsNoCase` with OrdinalIgnoreCase in `.qsearch` outputs

- Minor syntax cleanup


Kwoth:
I don't like embed support. I saw qsearch as something that find the quote by its content and prints the raw output. So no for that part.

Lower and upper string search seems dirty, there should be a way to compare them by ignoring the case.

shikhir-arora:
Done. This is the most efficient implementation for the general case here which is going to be like that for quotes. Also removed the embed as you asked. If you want to add a check for me to *require* the two be of the same length (personally I think this is fine without the extra overhead of doing that), I can add ` && q.Text.Length == text.Length` but that is another `==` operation that isn't really needed IMO. 

This should be a clean and fast way of getting what we want with a good flexibility margin :)

src/NadekoBot/_Extensions/Extensions.cs
```diff
@@ -278,6 +278,15 @@ public static IEnumerable<T> Shuffle<T>(this IEnumerable<T> items)
                 return list;
             }
         }
+    
+        /// <summary>
+        /// Easy use of fast, efficient case-insensitive Contains check with StringComparison Member Types 
+        /// CurrentCulture, CurrentCultureIgnoreCase, InvariantCulture, InvariantCultureIgnoreCase, Ordinal, OrdinalIgnoreCase
```
shikhir-arora: Technically, the valid options here to match our ContainsNoCase 'label' would be the CaseInsensitive StringComparison overloads: namely the three  CurrentCultureIgnoreCase, InvariantCultureIgnoreCase, OrdinalIgnoreCase

PR #1103, opened 2017-03-02
Music module changes

samdivaio:
Nadeko all time music emoji is added back again.
Queuing playlist now live updates remaining songs to queue.
Enable = Green
Disable = Red
Nothing major changed.


Kwoth:
No. Succes = green, Error = Red. Both enable and disable are success. I thought about this before.

samdivaio:
Same but then I saw removed songs so I thought it's probably not the same for music module. Want me change just that back to success green, failure red?

Kwoth:
Nah don't waste effort, I have another idea of how to adds emojis for module before the response in the future.

PR #1125, opened 2017-03-13
Installer Optimization 

samdivaio:
- Can copy instead of downloading the libs for x86 Windows users.
- Will fix the folder name white-space issue.


Kwoth:
thx sammy!

PR #1127, opened 2017-03-13
Don't exclude bots from the party.

nickhudkins:
Hardcoding the ignoring of bots is obnoxious. Treat them like users, ban them if they misbehave, don't grant them permissions that you don't want a person having.

### DIFF
--- a/src/NadekoBot/Services/CommandHandler.cs
+++ b/src/NadekoBot/Services/CommandHandler.cs
@@ -233,7 +233,7 @@ namespace NadekoBot.Services
             {
                 try
                 {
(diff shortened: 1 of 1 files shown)


samdivaio:
b1nzy gonna be super proud 😄

nickhudkins:
@Kwoth closed because....? There is an entirely valid use case here. If you'd like I can make it configurable instead, but this seems like one of the more silly checks to have.

Kwoth:
Closed because no, nadeko will not receive commands from other bots.

nickhudkins:
Sure it does, just run my fork @Kwoth.

PR #1141, opened 2017-03-21
A slight edit on replacing OwnerID

DogeOps97:
A slight edit on replacing OwnerID 
If people asks about it, then it's not clear enough(?) lol http://i.imgur.com/WxZiu3G.png


DogeOps97:
Oh wait what o_o

PR #1157, opened 2017-03-28
Added: View/search quotes by quote ID (.qid [id] / .quoteid [id]) 

shikhir-arora:
Per the request of @ManuelFte earlier today 🙂
I liked the idea he requested, so here we are:

---
## Description

- Added a `.qid [num]` or `.quoteid [num]` search option that allows you to search by the quote ID's that were recently put in. 

## Motivation and Context

- @ManuelFte asked me earlier if I thought it was a good idea and I said it sounds fine, and implemented it :-) 

- But really - makes looking up quotes that you want from the index with the ID good. We can delete them already, so it makes sense to look them up as they are with the ID.

- *This does support embeds on output* unlike `.qsearch` which was designed then to be a search-and-find "raw" utility, which it works well for since one can view their embed code from Discord if they need to for some reason instead of the database. For this, we want to just see the original content/quote, as many times there are 5000 of the same keyword and this serves as a lookup moreso than a raw output search. 

## How Has This Been Tested?

- This is pretty self explantory for this PR, as explained above.

- Usage is `.qid [id]` `.quoteid [id]` and the output display is similar to `.qsearch` but with embed support, a few minor changes to the text but same exact idea. See screenshot below.

## Screenshots

qid search example

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My change requires a change to the documentation. (✔️)  [autogen next update]
- [x] I have updated the documentation accordingly. (✔️)
- [x] I have added tests to cover my changes. (✔️)
- [x] All new and existing tests passed. (✔️)


shikhir-arora:
Oh, the spelling for success for the `if (success)` etc. was spelled sucess or something, so I just changed that (was on the DeleteQuoteById) 👍

Kwoth:
'Error' being outputted as an error message when quote isn't found is not enough.

Add a new string to responsestrings and use that key.

Kwoth:
Can you update this?

shikhir-arora:
Sorry this is messy, will open new PR.

PR #1178, opened 2017-04-11
Added: Show list of quotes that match a given keyword (ordered by quote ID)

shikhir-arora:
## Description

- Added a `.liqukwd [keyword] [optional page num]` search option that works like `.liqu` but lists quotes *only* matching a specific keyword, in order by quote ID (so approximately by when it was added)

## Motivation and Context

- The `.qid [quoteID]` command added earlier that shows the quote by quote ID works, but to find the quote ID `.liqu [page]` can be very large; often one wants to search under a matching keyword to narrow it down. 

## How Has This Been Tested?

- Self explanatory for this PR, as explained above.

- Usage is `.liqukwd [keyword]` or `.liqukwd [keyword] [page num]` and the output is a list similar to `.liqu` but ordered by quote ID (so roughly in added-order) for a specific keyword.

- Keyword is case-invariant (converted to uppercase to match database prior to searching)
## Screenshots

liqukwd search example

liqukwd help menu

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My change requires a change to the documentation. (✔️)  [autogen next update]
- [x] I have updated the documentation accordingly. (✔️)
- [x] I have added tests to cover my changes. (✔️)
- [x] All new and existing tests passed. (✔️)


Kwoth:
no, no more quote commands

PR #1216, opened 2017-05-10
Fix for pokemon trivia

DogeOps97:
Changed the `.xyz` to `.me`


Kwoth:
well thx but i'm blind, pushed it myself

PR #1232, opened 2017-05-31
Tag filter for ~e621

coggy9:
The fork my Discord server uses has a tag filter for the e621 command that blocks anything with tags that are against the rules or do not show up in the embed, like flash files. 

This adds filtering without adding to the tags that the bot has to search, as its limited to only six by the site.

A feature request probably needs to be filed for this, but it would be nice to see someone improve and expand this to include more NSFW searches and have the tags configurable. As of right now I'm just keeping the fork up to date with public NadekoBot.

### DIFF
--- a/src/NadekoBot/Modules/NSFW/NSFW.cs
+++ b/src/NadekoBot/Modules/NSFW/NSFW.cs
@@ -255,12 +255,26 @@ namespace NadekoBot.Modules.NSFW
                 using (var http = new HttpClient())
                 {
                     http.AddFakeHeaders();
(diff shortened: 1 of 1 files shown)


Kwoth:
Thanks for showing interest in contributing <3 

However, huge changes to the code are currently occurring in the 1.4 branch which are not yet pushed as I'm fixing the last few things. Another thing is that if this was done, it would have to be configurable on the module level (for all commands), not just for a single one, and not hard-coded but stored in the database.

This has been requested multiple times, and I'll consider it sometime, but atm it's very low priority.

PR #1246, opened 2017-06-09
Update Upgrading Guide.md

xnaas:
Since 0.9.x shouldn't even be functional anymore, I removed it and updated it for migrating from 1.3x and older.


xnaas:
Feel free to update anything I might have wrong?

PR #1247, opened 2017-06-09
Update Upgrading Guide.md

xnaas:
Touched up some language.


xnaas:
It still said 1.0...we're in 1.4 and some people might not upgrade for many, many versions. 😛

PR #1270, opened 2017-06-19
Update HangmanGame.cs

xnaas:
◯ is a really confusing character to have in the middle of words. _ would be far better.


xnaas:
img

Kwoth:
reason it's not like that, is because it bugs across devices. Mobile users reported weird issues with the text.

xnaas:
@Kwoth that was a long time ago. I can show you it's fine on mobile devices now too. 😛 

The original problems was that on mobile _ inside of embeds or `this kind of text` was still being italicized. That's no longer a problem as far as I can tell.

xnaas:
On mobile:

mobile

Looks basically identical to desktop. 😛

PR #1384, opened 2017-07-11
Fix error in Music command docs

iPonyRadio:


Kwoth:
that's not how it works, i'll update it though, thanks

PR #1403, opened 2017-07-13
Add deny perm for addReactions on mute

hakufu:
Added the permission for the channel override when muting to also include the denial of "addReactions." This is necessary since when muting it previously overwrote all channel perms(so it doesn't miss new channels,etc.) and there's no point in muting if they can still abuse reactions as well.


Kwoth:
Cool, thanks

PR #1513, opened 2017-08-18
Implement Auto NSFW from FeatHub

halitalf:
My implementation of [Auto NSFW](http://feathub.com/Kwoth/NadekoBot/+30). I also gave NSFW.cs some organization:
 1. Internals (in a region)
 2. Public aliases
 3. everything else

Tested `.autoboobs` and seems to work fine. No warnings on build/run.


halitalf:
As a side note, I was unable to create some translations on my own. Namely the following:
 1. he-IL
 1. ja-JP
 1. ru-RU
 1. sr-cyrl-rs
 1. zh-CN
 1. zh-TW

Kwoth:
Thanks

There are few things missing though
- Remove all translations except the main entry in en-us, they will be translated by translators
- Why are they exclusive? You should be able to have them both enabled
- Command description doesn't say you have 20s minimum limit

halitalf:
@Kwoth ok, I'll remove them now. I figured that if all three were running it would be very spammy, but OK. I can change that.

Issue #1554, opened 2017-09-03
Logs are restricted to having to be both bot owner AND administrator


Blacnova:
My friend invited my selfhosted nadekobot on his server, and wanted to enable .antispam and found that it doesn't tell you when you get muted. So I told him he could enable logging, but when he went to enter .logevents he was told that he had to be bot owner. Same thing happened with .log.

Kwoth:
This is more of a feature request. Please post it on feathub. Logging was disabled because of public nadeko.

PR #1555, opened 2017-09-04
Fixed Log Permissions

Blacnova:
Took care of #1554 just to make life easier.


Kwoth:
No, I have to take a look and see if it can be allowed on public nadeko.

PR #1602, opened 2017-09-16
Re-phrase, etc.

shivaco:
Re-phrased some lines, added gif for using `.permrole`, fixed the mistypes, links.


Deivedux:
I don't like the looks of some changes, might need to localize a bit more :/

PR #1632, opened 2017-09-26
Change "atleast" to "at least" in .clubcreate explanation

numbermaniac:


Kwoth:
1.4 is no longer maintained.
Also, don't update command list readme file. It's automatically generated with .hgit command

PR #1682, opened 2017-10-04
Fixed some polish translate

SebeCK73:


Kwoth:
This is not how it works. There are polish translators on poeditor, and they're gonna fix it, thanks anyway.

PR #1683, opened 2017-10-04
Disable die for StartupCommands 

shikhir-arora:
Just a small PR, if you wish. The `.die` command poses a risk as a startup command for self-hosters/botowners and shouldn't be allowed. It can cause the bot to lock. This small bit can also be used for other commands as it's just the same regex logic. 

There are four regex groups, `0-3`; `0` always returns the original data (so the command itself, say `.die`, `1` will always return the prefix of the command no matter its complexity, `2` will always return the text we are capturing as part of the regex - in this case `die` - and `3` will always return the remainder of the command or an empty string. 

This first checks to see if the command has `die` in it after the prefix, and **only** if it begins with die (regardless of the command prefix) will it continue to check the regex group `3` which *always* returns the remainder or an empty string. 

If it returns an empty string, then we can be certain that it's not another potential command that starts with `die` -- pretty straightforward.


Kwoth:
Good thing, but I can't accept PR now since I'm working hard on 2.0 locally. I'll make the change myself tomorrow in 2.0 and credit you in the code.

shikhir-arora:
No problem. :)

PR #1692, opened 2017-10-07
sucesful -> successful

shivaco:


Kwoth:
I've changed it on poeditor, will be fixed next update

PR #1746, opened 2017-10-24
Added redis, which disappeared for some reason...

MaybeGoogle:


Kwoth:
because you're looking at teh wrong branch :D 2.0 is a dead branch, not used anymore, lemme delete it thx

Issue #1781, opened 2017-10-30
Spelling errors in hangman


NogiKAZA:
I got notified by some on my server members, about spelling errors in the hangman game. 
Showd in the picture down below. There is probably more, but those were the one we managed to find. Thanks.
![image](https://user-images.githubusercontent.com/33229759/32181920-d1cdf614-bd95-11e7-82c9-2eefd1bb7c5a.png)

Kwoth:
xnaas PRed a fix. If you're on windows, wait until next release, if you're on linux, just update right away.

PR #1784, opened 2017-10-30
Fixed Typos?

xnaas:
I hope editing this file is good enough to solve #1781? Sucks that the URLs still have to be literally wrong...but whatever!


Kwoth:
tyvm babe

PR #1805, opened 2017-11-07
force google search results to english.

MaybeGoogle:
gws_rd=cr,ssl →google web server_redirect=country,ssl(i have no clue what the ssl is for...)
I've tried replacing cr with `us` and `ncr` (no country), but these dont do anything. 

This guarantees that search results are converted to english: https://www.google.com/search?q=test&hl=en

Tested with tor set to Austria.


MaybeGoogle:
converted to english = english/united states results, not translated. In case that couldve caused some misunderstanding.

Also, this way opens up the possibility of different languages being supported on google search command...

shikhir-arora:
I think the idea here is that most selfhosters will host in their country (many just host it on their PC's), so the `gws_rd=cr,ssl` tag will ask Google Web Server to redirect using the country (the IP of the place it's hosted) and SSL simply means to use the encrypted (`https://`) version of Google Search (forcing it).

The English thing may be a good idea for public Nadeko, but probably the way it is right now is best for the general source.

MaybeGoogle:
@shikhir-arora thanks for the ssl clarification.

The thing is on public nadeko, google results are being displayed in german. There is no really easy way to fix this. Instead, I think there should be a list of languages that will replace the language (en in this case) with the language that corresponds to the langset of a server. 

I agree that this pr is just a temporary fix. A more sound fix would be something like what I described above.

MaybeGoogle:
changed the branch to 2.0 cuz it looks like you're using that again.

shikhir-arora:
Pretty sure 1.9 is the main branch, 2.0 is just dev work and it gets merged into 1.9 anyways.
Also pretty sure Kwoth is keeping at 1.9 for now because the bash script is still setup to clone from 1.9 so it's easier that way. 👍

Kwoth:
2.0 is not maintained.

Issue #1366, opened 2017-07-09
Bot after youtube-dl update will not display Japanese/Chinese characters in song information.


OrigamiSS:
After the new update the now playing and queue list will show random symbols instead of actual characters . This is repeatable only on windows version of Nadeko bot.

https://cdn.discordapp.com/attachments/330949870816854027/333397973847375873/image.png

ManuelFte:
In my case, the Windows install ignores any non-Latin character (meaning a song with only Japanese/Chinese characters in the title will appear as having no name). The Linux version doesn't have this issue.

Kwoth:
I actually have no idea how to solve this.

samdivaio:
python issue? https://github.com/rg3/youtube-dl/issues/13185

willysunny:
Just a heads up, I can NOT reproduce it on docker with Nadeko V1.7.1 anymore. So youtube-dl might have fixed it themselves?

samdivaio:
NadekoBot 1.5.1 doesn't even use YouTube-dl, its introduced with version 1.54a I guess.
Also the issue is related to Python and it can be easily fixed on linux https://github.com/Kwoth/NadekoBot/issues/1392 docker already uses the locale required.

Macley-Kun:
So can we confirm that the issue is still present lads? If it's fixed then i geuss the issue is no longer relevant.

Kwoth:
it is

Macley-Kun:
I suggest we could add the help wanted label on this? Maybe someone knows how to resolve this!

Macley-Kun:
Because we need a similar answer in: #1455.
I'll mention the issue on both, so if someone would be able to help, he/she can also help with the other issue.

willysunny:
So ... correct me if I'm wrong. So basically this is just a local issue?

But I don't see a locale-gen in youtube-dl for windows version?

Macley-Kun:
Correct me if i'm wrong. But the youtube-dl isn't something Kwoth can edit/change right? Wouldn't it be better to open an issue on Youtube-dl Github? We could then refer to that issue from here and then close this one (or close it when they add support for it).

Kwoth:
I think it should work, as youtube-dl is only used as a fallback now.

PR #1817, opened 2017-11-11
fix crca wont work on chinese sentences

coconut24:
getwordposition should ignore special characters, but still look for valid word chars(A-Z, a-z)
should i make it look for characters below =<255? (eg. "@{}()" and such) :thinking:

### DIFF
--- a/NadekoBot.Core/Modules/CustomReactions/Extensions/Extensions.cs
+++ b/NadekoBot.Core/Modules/CustomReactions/Extensions/Extensions.cs
@@ -133,15 +133,15 @@ namespace NadekoBot.Modules.CustomReactions.Extensions
 
             if (indexOfWord == 0) // on start
             {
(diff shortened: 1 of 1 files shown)


Macley-Kun:
Thanks for making this pull to fix our issue we had for about 3 months!
Idk if its possible to have a look at This issue: #1366. Its Kinda simalir But perhapes u can make a pull for this one too. Still thanks again for making this pull :33

Kwoth:
Thanks a lot, but yeah, some special characters which are used often, like @ and #, etc could make a problem, if you had a crca with them. Not sure who would put them as CRCA expecting the sentence to contain a word "?", i think it's fine like this?

Kwoth:
#1817

coconut24:
another way to fix this is detecting cjk characters, but korean does use spaces. and theres many languages that doesnt use space as word divider or word divider at all.
it would be unnecessary to list all languages without/with different word dividers

Kwoth:
Thanks for the help

PR #1825, opened 2017-11-12
:'( redis still not there

MaybeGoogle:
hope this is the right place this time


MaybeGoogle:
maybe a nadekoOSXautoinstaller.sh with all of those prereqs cuz the list is getting too long for me to type all of them ;)

PR #1838, opened 2017-11-16
Allow for optional "titleurl" field for embed title [clickable]

shikhir-arora:
This allows the title to be clickable from a long-standing iOS bug.

It is optional, but when inputted the `"url": "http://something.com"` will act as it should and make that field clickable. 

`{ "title":"Something", "url":"https://google.com" }` (with URL for the title)
`{ "title":"Something" }` (without URL, as normal) 

Have tested and confirmed working. Pictures are on the commit below.


Kwoth:
The field has to be called just `url`, in order to be inline with api and discord.net model.

shikhir-arora:
Changed to `url`

Kwoth:
You also have to check whether the URL is valid, see how thumbnail and image are done.

shikhir-arora:
Should be good to go.

Kwoth:
ty

Issue #1818, opened 2017-11-11
Nadeko changing role perms for custom mute role


tukimoop:
I believe Nadeko wasn't supposed to change channel role perms on mute roles that were set using `.setmuterole`, as mentioned by Zerg and having seen a 1.8 version of the bot before not changing channel roleperms with custom mute roles. However, on 2.3.6 (and presumably a few more versions before this one), the bot would change channel perms whenever `.mute` was passed.

Role perms changing:
![admin](https://user-images.githubusercontent.com/28344254/32689004-75726992-c72f-11e7-9206-d6d22f528421.gif)

Intended vs unintended effect of what most people want:
![user](https://user-images.githubusercontent.com/28344254/32689006-8473f5fa-c72f-11e7-8e26-75019d463307.gif)

Kwoth:
fixed in 2.5.6

PR #1841, opened 2017-11-17
.mute shouldnt overwrite existing mute-role channel permission

coconut24:
GetMuteRole will only add/overwrite mute-role channel permission if theres no channel permission targeted to the mute-role
this should fix https://github.com/Kwoth/NadekoBot/issues/1818

### DIFF
--- a/NadekoBot.Core/Modules/Administration/Services/MuteService.cs
+++ b/NadekoBot.Core/Modules/Administration/Services/MuteService.cs
@@ -1,17 +1,16 @@
-﻿using System;
-using System.Collections.Concurrent;
-using System.Collections.Generic;
-using System.Linq;
-using System.Threading;
-using System.Threading.Tasks;
-using Discord;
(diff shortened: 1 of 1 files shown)


NadekoBot.Core/Modules/Administration/Services/MuteService.cs
```diff
@@ -192,7 +192,7 @@ private Task Client_UserJoined(IGuildUser usr)
             {
                 try
                 {
-                    if (!toOverwrite.PermissionOverwrites.Select(x => x.Permissions).Contains(denyOverwrite))
+                    if (toOverwrite.PermissionOverwrites.Select(x => x.TargetId == muteRole.Id).Count() == 0)
```
Kwoth: maybe `toOverwrite.PermissionOverwrites.All(x => x.TargetId != muteRole.Id)` ?

Issue #1747, opened 2017-10-24
Double sending of messages in v2.1.2 on bot 


Radioboy126:
I've hosted Bots on Digital ocean and had backed up my former bot's info and now the bot is double sending messages. I am running one instance of the bot
![doubleshard](https://user-images.githubusercontent.com/32051740/31929023-221988be-b84f-11e7-8ce3-d3e0a8362ec7.png)
![single instance](https://user-images.githubusercontent.com/32051740/31929032-24851bf4-b84f-11e7-9dce-356529c7dc8a.png)

Kwoth:
Update

Kwoth:
Update again, i pushed a possible fix.

Radioboy126:
@Kwoth uh... i think it didnt http://prntscr.com/h19eqz

VsFp:
hi, i updated mine but it still shows 1.9+ what can i do?

Midoriin:
@Kwoth This is an issue caused by the automatic shard restart system, I guess. Today I saw a startup command message even though shards were fine. Then I started to get double responses.

MaybeGoogle:
Happened to me when I updated to 2.1.2. No problem in the beginning, though, when I just started running the bot. 

However I looked at the console and while the startup commands (only executed one time btw) and the "doubled" commands are there, the ping and custom reaction (also only one response for each) that I manually executed just as the bot started up are not displayed :thinking:

MaybeGoogle:
No longer happening to me, dont know about anyone else. v2.3.2

shikhir-arora:
I was never able to replicate this. Only thing I can think of that caused this for some users is if Nadeko wasn't properly shut down prior to update.

You can do something like: `ps aux | grep Nadeko` and will likely see something like this:

```root      5333  0.0  0.0  12644  3188 ?        Ss   00:58   0:00 bash /home/vlexar/main/vlexar/NadekoARN.sh
root      6118  0.1  0.1 3341764 60792 ?       SLl  00:58   0:01 dotnet exec /home/vlexar/main/vlexar/NadekoBot/src/NadekoBot/bin/Release/netcoreapp2.0/NadekoBot.dll
root      6323  1.7  0.5 4089568 166832 ?      SLl  00:58   0:20 dotnet exec /home/vlexar/main/vlexar/NadekoBot/src/NadekoBot/bin/Release/netcoreapp2.0/NadekoBot.dll 0 6118
[...]

PR #1863, opened 2017-11-20
Fix for many of the 'double' spawn instances (reference #1747) [0/1 shard bots]

shikhir-arora:
**[This will work for anyone using a 0/1 shard bot, as that's all this tests for]** 

*I am sorry this is a long explanation but I wanted to be clear about it for others potentially reading.* 

So after a **lot** of going back and forth on #1747 - I was finally able to replicate the issue reliably. I used `.restartshard 0` to restart the shard, and it came back with two instances. Do it again and there were 3, etc. To make a long explanation short (much of it is in #1747 -- redis persists the data as it's still running after Nadeko restarts a shard. Even if you change the ShardsCoordinator to push a delete (like it does on `.die`) for the  `shardcoord_stats` table, that doesn't stop it. *And*, even after COMPLETELY killing every dotnet process and script, checking this multiple times, and restarting -- Nadeko would spawn an instance more than the last. Nothing worked. So there was an issue here. But:

### **tl;dr** -- It persists in redis, and the only way to kill it is to kill the running process for redis which you should do after stopping Nadeko (as Nadeko is writing to redis live) and then it will 'reset' and the restart will be smooth, as mentioned on my comments here: https://github.com/Kwoth/NadekoBot/issues/1747#issuecomment-345664265 

That explains most of my findings. But that fix of redis restart isn't a fix, only temporary, so the issue persists.

---

This seems to fix it. And I've confirmed it with doing **everything** I can many, many times. For example, I've done `.shardrestart 0` and it cleanly restarts the shard. I watch this with two console windows monitoring the dotnet processes and my pm2 process monitor which shows the realtime Nadeko logs, startup, etc. You will see that it will terminate and restart the shard properly. 

Now that **does not** confirm everything -- namely the ShardsCoordinator doing the auto-restart job without the extra spawns, so I went ahead and replicated a force-shard restart (from the ShardsCoordinator) - here's a normal output of `ps aux | grep dotnet`:

```sh
vlexar@node1:~/main/vlexar/$ sudo ps aux | grep dotnet
root     29449  2.2  0.2 3364388 78284 ?       SLl  12:50   0:00 dotnet run -c Release
root     29646  2.8  0.2 3349780 72940 ?       SLl  12:50   0:00 dotnet exec /home/vlexar/main/vlexar/NadekoBot/src/NadekoBot/bin/Release/netcoreapp2.0/NadekoBot.dll
root     29665  2.6  0.2 3364368 78660 ?       SLl  12:50   0:00 dotnet run -c Release -- 0 29646
root     29826 30.1  0.4 4171568 141436 ?      SLl  12:50   0:04 dotnet exec /home/vlexar/main/vlexar/NadekoBot/src/NadekoBot/bin/Release/netcoreapp2.0/NadekoBot.dll 0 29646
vlexar   30233  0.0  0.0  14244   996 pts/3    S+   12:50   0:00 grep --color=auto dotnet
```

Processes 29665/29826 are the child processes that are normal which are the ones that mimic the Shard. I went ahead and `kill -KILL 29665` / `kill -KILL 29826` to "force" Nadeko to act like it was restarting a shard. 

---

The bot went unresponsive to my commands (as it's a 0 shard bot so 100% normal/expected) for a few seconds.....but, after the proper 7.5 seconds or so (per ShardsCoordinator):

```bash
1|Nadeko   | -1 12:52:13 ShardsCoordinator | Shard 0 is scheduled for a restart because it's unresponsive.
1|Nadeko   | -1 12:52:19 ShardsCoordinator | Auto-restarting shard 0
```

And it came up 👍  I repeated this test **many** times, waiting different times in between. And I also did a few other ones: restarting Nadeko with `.die` then doing a `.restartshard 0` and *then* killing the child process to mimic the shard going down. And it does actually work *very* well, just like I'd expect. When I do the manual kill Shard 0 goes down, the bot is unresponsive for just a few seconds -- again, as it should be, but it comes back every time after a few seconds with the scheduled auto-restart. 

---
And there is nothing that is active and running that's consuming resources that are leftover; the process is killed with `Environment.Exit(0);` and confirmed to work. 

It's just two lines of a fix, so I apologize  for the long explanation, but I wanted to be clear about how I came to this. This wasn't even an issue that affected me, but after a lot of trial and error I was able to replicate it and then come to this.

### DIFF
--- a/NadekoBot.Core/Modules/Administration/SelfCommands.cs
+++ b/NadekoBot.Core/Modules/Administration/SelfCommands.cs
@@ -276,6 +276,8 @@ namespace NadekoBot.Modules.Administration
                     JsonConvert.SerializeObject(_client.ShardId),
                     StackExchange.Redis.CommandFlags.FireAndForget);
                 await ReplyConfirmLocalized("shard_reconnecting", Format.Bold("#" + shardid)).ConfigureAwait(false);
+                await Task.Delay(6000).ConfigureAwait(false);
+                Environment.Exit(0);
             }
 
             [NadekoCommand, Usage, Description, Aliases]


coconut24:
wont this only fix restartshard?, i think shardscoordinator will try to kill the shard when its unresponsive, if not properly

shikhir-arora:
@coconut24 This will work on any **single-sharded** bot. At least that is all it was tested/meant for but it could work on many shards, I just can't test that. It's not really a 'permanent'  fix; my comments on that and more are linked here: https://github.com/Kwoth/NadekoBot/issues/1747#issuecomment-345664265 

**If you wish to try this out**, simply delete the `SelfCommands.cs` file from `NadekoBot.Core/Modules/Administration` and:

 `wget https://raw.githubusercontent.com/shikhir-arora/NadekoBot/1.9/NadekoBot.Core/Modules/Administration/SelfCommands.cs` 

**Make sure** your bot is dead/offline and start it up. If you are running the latest version right now, and type `.restartshard 0`, you'll see it boots up with two instances. You can also try to `.die` and that will probably boot another instance. Or another `.shardrestart 0` and it'll do another. Restarting it normally won't help. This can be verified. And, to answer your question directly, I did address that in the second half of this PR. See the part where I address that. Yes, I've killed the child processes and the ShardCoordinator will restart them properly, **without any issue.** This does work with that as well. I've also had a few others test and confirm this works. 

You can mimic killing the shard like I did in the child process explained above. This will kill the child process, and the bot will (as it should) go unresponsive. You'll see ShardsCoordinator restart it properly. Why does it actually work? It's sort of a trick that tells tells the main process to "shut off" after a delay (6000ms here) upon the bot restarting the shard. All of this is actually in redis (`redis-cli --slave` for live output) and basically this will cause the following loop:

- For some reason, be it `.restartshard` or killing it (so the shard being unresponsive), ShardsCoordinator will try to restart the Shard
- The code here will continue this as usual, but kill the process that it actually initiated. It's actually the parent killing itself *but* when it restarts the "extra" instance that is caused by the bug in ShardCoordinator takes the shard 0 position and becomes the parent process again. This is how it continues, each time cleaning the "garbage" of the prior. Not the most elegant fix...but it works for 0 shard (1 shard) bots perfectly.

- The extra process will be killed and the new shard will restart. The delay ensures they don't collide. Remember, the way ShardCoordinator is right now, it creates a new process, restarting the bot completely. This kills off the old parent process but replaces it instantly with the new instance. Again, please feel free to try it out if you wish as it's just two lines. If you find it working/not -- let me know!

- The actual bug should be looked at by @Kwoth closer. It looks to be a clear issue with stopping in Redis, as I've observed this live, and persistence. Something like `WaitForExit()` would be a starting point perhaps for the `Process.Start` in ShardsCoordinator. However, this can be thought of as a very quick/easy/harmless fix for most users and can be overridden or reverted anytime.

coconut24:
thanks for the tutorial, if the old shard still persist, it could be that shardcoordinator failed to kill the old shard. this method wont work if the shard is frozen/unresponsive, but the new shard will still work anyway

shikhir-arora:
Like I noted it’s intended for most users (99+%) who use one shard. That’s technically no sharding and so this will work fine for those, which are the majority of users having this issue. It’s not tested on multi-shard bots and it’s safe to say that in probably wouldn’t work for them. However, my main motivation for this was to get people a temporary fix because we see tens of people complaining about this issue and posting here. I also wanted to post my findings and how I came to the convulsions I did. But I’m sure a permanent fix is in the works.

coconut24:
found out the problem was dotnet having a child process to run the executable
screenshot_2
screenshot_3

which will free the child process when shardcoord tries to kill the parent

coconut24:
i can confirm this method will work https://github.com/Kwoth/NadekoBot/issues/1747#issuecomment-346250181
natural restart still causes problem, but can be temp fixed with the pr'ed .restartshard
ang thats kinda ineffecient. maybe you can find perm fix for this as i dont have much time

coconut24:
i can confirm this method will work, check https://github.com/Kwoth/NadekoBot/issues/1747#issuecomment-346250181
natural restart still causes problem, but can be temp fixed with the pr'ed .restartshard
and thats kinda ineffecient. maybe you can find a perm fix for this as i dont have much time

Kwoth:
Just pushed a possible fix, I tested it, seems to work fine.h

shikhir-arora:
Should be fixed then via e2ddc3c

PR #1868, opened 2017-11-22
Too many people forgetting it

MaybeGoogle:
Comma in credentials


coconut24:
instersting... cant you tell people to add these two lines before the end bracket and put a comma before the first line instead? json would be able to parse that normally

shikhir-arora:
I think the comma issue was addressed in https://github.com/Kwoth/NadekoBot/pull/1870

MaybeGoogle:
@shikhir-arora no, that looks like a different line.
@coconut24 or even better, put those two right after the open bracket and add a comma on the second line. That way formatting is maintained. thanks!

Kwoth:
Thanks

PR #1870, opened 2017-11-22
JSON Explanations Updated

samdivaio:
- Rearranged the example details according to the `credentials_example.json`
- Added comma to the restart function to prevent issues if they are copied.


Kwoth:
ty sammy <3

PR #1882, opened 2017-11-27
fixes double shard on restart

coconut24:
as far as i know, dotnet run will create a sub/child process to execute the executable (dotnet exec).
which may leave the child/shard running, as shardcoordinator only takes and kills the parent process (dotnet run).

in this pr, i found a way to kill the process tree (starting from chosen parent) which would be able to completely kill the running shard, on both unix and win platform.

i also removed the killing of the shard in _shardStartQueue watching task, as it shouldve been handled and stopped by OnShardUnavailable method. i dont see any problem on redis's end


coconut24:
oops i think i removed trypeek :P, tell me if you want it to be added again

Kwoth:
Thanks a lot for this.

Yea trypeek needs to be there in order to avoid race condition, where the id gets dequeued, starts the shutdown piece of code, and while doing it, gets queued again by the disconnection message logic.

coconut24:
done :)

coconut24:
done :^)

Kwoth:
Are you sure this works? :D

coconut24:
cant you try it out >.>
it worked for me on both win 8.1 and ubuntu

coconut24:
cant you try it out >.>
it worked well on both win 8.1 and ubuntu

NadekoBot.Core/Services/ShardsCoordinator.cs
```diff
@@ -228,11 +234,7 @@ private void OnShardUnavailable(int shardId)
                         {
                             _log.Warn("Auto-restarting shard {0}", id);
                         }
-                        var p = StartShard(id);
-                        var toRemove = _shardProcesses[id];
-                        try { toRemove?.Kill(); } catch { }
-                        try { toRemove?.Dispose(); } catch { }
-                        _shardProcesses[id] = p;
+                        _shardProcesses[id] = StartShard(id);
```
Kwoth: We have to make sure to kill the previous one, if there's already a shard with that id running

coconut24: @Kwoth onshardunavailable wouldve killed the process. doing twice might cause problem, itll kill the same pid twice. also there wouldnt be any use for it if its already nullified right?

is there a case when restarting a shard without being handled with onshardunavailable?

Kwoth: Not atm i think, but there might be, just do the same thing as in other place please, just check for null, and if it's not kill it, can't really cause the problem.

coconut24: im currently on a vacation, feel free to push changes to my branch, otherwise just copy changes to the main branch if you dont feel like doing it.
thanks

shikhir-arora: Can confirm this works fine with the removed few lines inserted back. 👍 

```cs
var p = StartShard(id);
var toRemove = _shardProcesses[id];
try { toRemove?.Kill(); } catch { }
try { toRemove?.Dispose(); } catch { }
_shardProcesses[id] = p;
```

Kwoth:
Thx, merged

PR #1887, opened 2017-11-30
Included download link, word fix, and style fix

willysunny:
Since docker decided to remove the toolbox main page from their site, I changed the download link to direct link.

Also taking the suggestion from the previous edit.


willysunny:
Because it's not for mobile?

PR #1898, opened 2017-12-05
added 17.10, forgot. Thanks vlexar

MaybeGoogle:
clicking buttons is good exercise, i heard


Kwoth:
i'm a body builder then.

PR #1908, opened 2017-12-08
Path of Exile Related Commands

SimonNguyen:
Multiple Path of Exile related commands:

- .poe \<Account Name\>: Returns a list of character information on the account, if account is public. Paginated, 10 characters per page. https://i.imgur.com/gL2pUlX.png
- .poec \<League Name\> "\<Currency Name\>" "\<Desired Currency Name\>": Returns the exchange rate between the given currency and desired currency. Desired currency is optional. https://i.imgur.com/g6JIQxA.png
- .poel: Returns a list of currently available leagues. These names are valid for the .poec command. https://i.imgur.com/jMHo9uP.png
- .poei "\<Unique Item Name\>": Returns the tooltip for the given unique item. https://i.imgur.com/LwcwWHo.png


Kwoth:
I like the commands a lot, thanks.

I'll check the code later, but the small problem I see right away is that you're not using the httpClient from the service, but creating your own. This causes some issues https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/

tl;dr Use a static httpclient from the service. If the service doesn't have one, add it.

SimonNguyen:
Interesting article, I was unaware of that since the docs themselves always had examples with using. The commands are now using the httpClient from the service.

Kwoth:
Very neat. Thanks a lot!

R4zzik:
Is .poei broken? the other commands seem to work just fine. Amazing work with the bot btw!

SimonNguyen:
Yes, the command broke after Gamepedia changed a Wiki module @R4zzik 

I plan on updating the command soon.

Kwoth:
@SimonNguyen any plans on fixing this?

BanhMiDacBiet:
@Kwoth using a different account now. I would recommend removing the item command. The wiki changed its API and I've been too busy to fix that.

PR #1914, opened 2017-12-10
pm2 support - README 

shikhir-arora:
This is the README for the PR here: https://github.com/Kwoth/NadekoBot-BashScript/pull/37 that enables pm2 automatic install and startup for Linux selfhosters as a seamless way to run Nadeko. Details are on that PR; this is the README as such.


Kwoth:
Hm, I don't like this being in the same file. @samdivaio what do you think? Should it be here or create a new entry "Linux with pm2" ?

shikhir-arora:
I’m fine with either by the way but it kind of makes things messy. Right now the logic depends on how it is, where you just need linuxAIO.sh and that is the master script. From there any option (say 6) installs pm2. It goes back to the linuxAIO script because the pm2 setup bash script deletes itself after it is done. Then you can exit, or choose - say option 7 - that takes you to the menu to start in pm2. Once you choose a startup there, it’ll delete that file. So it’s kinda all tied in. Sam wanted me to even make a master script  called linuxAIO.sh which downloads the main installer (the current linuxAIO.sh) in case of changes in the future — so users really just can have a potentially old version of the main script and it’s fine.

shikhir-arora:
I should probably elaborate. https://github.com/Kwoth/NadekoBot-BashScript/pull/37 you’ll see I made what Sam asked which is linuxAIO.sh and that downloads the master install script which is the current linuxAIO.sh + two options for pm2 install and startup. I think it makes sense because you can flow thru the script. So you run linuxAIO.sh, then choose 1 for Downloading Nadeko. Then it goes back - you choose 4 for setting the prerequisites. Then it goes back. Now you can choose a startup or do the pm2 setup and go back and pm2 startup. This script already serves the home central for everything else, so I don’t see why it shouldn’t here.

samdivaio:
If we create another entry for PM2 exclusively on the guide it might cause some confusion with the people who check out the linux part but unsure what the other two option actually do on the installer. Moving it to another section would be a lil inconvenient. @Kwoth its good to have it on the same place its gonna be lil more convenient and reliable with pm2. ❤

@shikhir-arora I believe the part on `line 48` needs to reflect the current installer options.

shikhir-arora:
@samdivaio Thank you for pointing it out, changes done.

shikhir-arora:
@samdivaio Just added some cleanup to make it clearer that it’s a continuation of the installer options, and also fixed 6 -> 8 as the exit option (referenced in the tmux section).

Macley-Kun:
Are there some things that need to be done before this get's pulled?

shikhir-arora:
@Macley-Kun No, everything has been checked on both PR's carefully for a while now.

Macley-Kun:
So i have tried out the pull you made and have some suggestion to improve the user experience more.
Using the script didn't cause any problems what so ever nor did i find it hard to use.

1. So when running the linuxAIO2.sh i noticed that you added 2 more options. This way is acceptable but i would suggest changing the following:

In `3. Run Nadeko with Auto Restart` Have the user be able to choose from:
**Run Nadeko in this session.**
**Run Nadeko with pm2.**

Optionally _after_ the option `3` is chosen. The user get's a message explaining what both options include. I would keep it short like: 

`You can choose between two options to run your NadekoBot. 
Running NadekoBot in this session means that you start the bot in the session you see now. You need to keep this session up to be able to keep NadekoBot working.

Running NadekoBot with pm2 means that you run NadekoBot in the background of your machine. You can close the session your currently using and NadekoBot continues to work.`
The user then continues to the 3 options.

When the user chooses option 1. Run Nadeko in this session. They get the `Welcome to NadekoBot Auto Restart and Update!` options that are already in NadekoBot.

When the user chooses option 2. Run Nadeko with pm2. They get the options 6 and 7 you use in your script.


2. I would maybe rename `6. Set up pm2 for NadekoBot (see README!)` to something like:
`6. Auto-Install Prerequisites (For pm2, see README!)`
And make `Start Nadeko in pm2 (complete option 6 first)` with a capital C. So it's all perfect :P...

3. Probably not a bug but maybe add in the command: `pm2 update` so people don't see any errors :).
image

4. It would be nice to name the App name to something a user wants to specify. For instance i use a bot called Delta. Some would prefer to name their app Delta.

TLDR:
Can you add an extra step in option 3. Let users choose between normal and the pm2 way. In option 3 if users choose the pm2 way, they get your options (6 and 7).
Might wanna add `pm2 update` to avoid In-memory PM2 is out-of-date error.
Add option to name your App the way you want?

Sorry for the long shitpost but hope it helps. It's just a suggestion so you can ignore this if you disagree. :D

Macley-Kun:
Also i'm having issues with using this script suddenly...
http://hatebin.com/akafetrsmp
For some reason the bot doesn't come up...

Gonna try to get it working with your script.

shikhir-arora:
@Macley-Kun 

Thanks, Macley. I’ll look at your suggestions and implement the ideas later today. I’m not sure why it doesn’t work that one time. But as **many** people have tested this including Sam and a few outside of the Nadeko support team, I’m sure it works. 

- I can add the `npm i -g npm` command sure. It doesn’t do anything but update npm and that message you’re seeing is just informational. But it doesn’t hurt, so sure.

- In terms of naming the bot what you want - that’s fine but for this guide it’s going to make things more confusing. I mean I *could* make a function that takes a variable and stores it and then passes the output to STDOUT to pm2 but meh. You can always rename a pm2 process like this: `pm2 restart <OLDNAME> --name <NEWNAME>` so I don’t think I’m going to add extra stuff to it now.

- Will fix the spelling and grammar things :)

A few other things - as noted in the documentation, this is **only for initial setup** of Nadeko. To update Nadeko afterwords you just want to `pm2 stop Nadeko` and use the download script (`linuxAIO.sh`) or guide from the website like it is now and just choose option 1, Download NadekoBot. Then `pm2 start Nadeko` and it’ll be updated. It takes about 15 seconds for me. If you compare it manually: much faster than attaching to tmux, stopping the bot, pulling the downloading, attaching back into tmux and starting up and then exiting / detaching.  Not to mention that 1) pm2 closes processes properly with SIGKILL and 2) pm2 manages containers and can be used for many things like uptime monitoring, long-term logging, and even load balancing.

Lastly, if anyone is trying this, don’t do it with *this* PR and the attached one. They’re set to download from Kwoth’s repo in the event that this is merged, so they won’t work properly. You have to contact me `.vlexar#5320` and I’ll give you a modified version, or change the links in the bash scripts to my local repo and not Kwoth’s 👍

Macley-Kun:
The problem was just my stupid actions 👀...
It's better to not add suggestion 4 of mines. If it's so easy to rename it i rather would add that 'optionally' in the docs.

Thanks for considering some of my suggestions 0/

shikhir-arora:
@Macley-Kun I implemented some of your suggestions. See: https://github.com/Kwoth/NadekoBot-BashScript/pull/37/commits/ccfee35b7100c05dfdd0184a8b5f647a8e0c8c07

Anyway, I think having the Auto-Restart like you mentioned is extraneous, but I added some language changes to make it more clear everywhere.

`pm2 update` isn't needed because it installs pm2 for you -- this would update it automatically.
Added the `npm i -g npm` to update npm just incase.

Thanks for the suggestions 👍

Macley-Kun:
Seems that pm2 has bin launched 👀 @Gremagol time to replace your shitty tmux

Gremagol:
Boy i dont care why do you mention me

PR #1918, opened 2017-12-11
Fix of aar help string

shivaco:


Kwoth:
Thanks, I actually saw that and forgot to fix it myself.

PR #1922, opened 2017-12-12
 Fixed calculation of VoiceChannel count

Anheledir:
Since categories were introduced, the number of voice channels was too large as it also contained the number of categories. (by Anheledir#6792)


Kwoth:
Makes sense, thanks a lot!

PR #1935, opened 2017-12-17
Added a simple .wants command

zergling-man:
Which shows who has affinity to you. Separated by which ones you own. No longer includes my token...
Also indenting is whatever, it's not Python.


Kwoth:
Can it get renamed to .waifuwants

zergling-man:
Is the alias fine?

Macley-Kun:
*bump*

Macley-Kun:
@Kwoth Any comments on this pull 👀 before were getting some upstream stuff...

shivaco:
Any news on this one? :thinking:

zergling-man:
What hasn't (until now) made it into this thread is @Kwoth's comments along the lines of "I don't see the point of having a new command for this". There was a brief consideration about adding it to `.waifuinfo`, though even that's apparently going to be a hard sell.
I think this information is extremely useful, personally, but I haven't come up with any better way of including it than the waifuinfo addition. I suppose I should at least do that and then see.

NadekoBot.Core/Modules/Gambling/WaifuClaimCommands.cs
```diff
@@ -471,6 +473,10 @@ public enum DivorceResult
                     await uow.CompleteAsync().ConfigureAwait(false);
                 }
 
+                // Dedupe, the right way.
+                affinities=affinities.Except(claims).ToList();
+                // Now it only contains unclaimed affinities.
```
Kwoth: this should be a part of the query, not using linq

zergling-man:
https://puu.sh/AwpbP.png - with an unclaimed affinity
https://puu.sh/Awpbw.png - with a claimed affinity. The * is not currently explained anywhere. I'm considering splitting that into two fields to be explicit about it, but I like the idea of it just showing up in the middle of the list. (The original intent of this is to allow you to see people you might want to claim, because they have affinity set. Seeing people you hold who have affinity or not is useful for considering divorces, but isn't something I explicitly care about supporting. Just figured it would be easy to do while I was at it.)

NadekoBot.Core/Modules/Gambling/WaifuClaimCommands.cs
zergling-man: You'll need to explain that in more detail to me. Keep in mind I'm learning not only Nadeko through field experience, but also dotnet.

Kwoth:
Wouldn't it make more sense to put a 💞 for claimed person with affinity towards you

Kwoth:
Can you change "Desires you" to "Loved by"

zergling-man:
Oh, both of those are great ideas. I'll do that promptly.

Macley-Kun:
@Kwoth 👀

Kwoth:
I'll consider this in 3.0

PR #1973, opened 2018-01-02
.localplaylst extensions fix

FredericaBernkastel:
I was wondering why bot tries to play `audiochecker.log` files...

### DIFF
--- a/NadekoBot.Core/Modules/Music/Music.cs
+++ b/NadekoBot.Core/Modules/Music/Music.cs
@@ -751,8 +751,11 @@ namespace NadekoBot.Modules.Music
 
             DirectoryInfo dir;
             try { dir = new DirectoryInfo(dirPath); } catch { return; }
+
+            List<string> validExtensions = new List<string>() { ".mp3", ".flac" };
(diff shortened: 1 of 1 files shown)


Kwoth:
Hm, i can't accept this if there is no comprehensive list of most commonly used formats which are playable by the bot. I bet a lot of people are playing .wav or other types of files, so the command will stop working for them.

FredericaBernkastel:
Maybe i'll add configurable option for extensions.
Just hint me, which files i should edit, because[ this is massacre](https://github.com/FredericaBernkastel/NadekoBot/commit/a7d78bcd56c88f766a7083e198cc79397284f922).

Kwoth:
Uh wow, idk what you did there. No, I don't want them configurable. Just a comprehensive list, we will add ones we've missed. But .mp3 and .flac are definitely not enough as a starting point

FredericaBernkastel:
I think we can start from `ffmpeg -formats` output.

Kwoth:
sounds like a plan

Macley-Kun:
Any update on this @FredericaBernkastel ?

Issue #1991, opened 2018-01-13
Nadeko Music....


NeeAfk:
I don't know what do anymore. I've reinstalled ffmpeg multiple times, each time restarting my pc. No matter what I do the music won't work. It had just been working when after a song ended it stopped the entire playlist and said **"Destroying"** in the cmd. After that I restarted the bot and now all I get is **"No song found"**. 
![image](https://user-images.githubusercontent.com/26068101/34901546-5284199e-f7d1-11e7-87fc-0499eaa0ec59.png)
![image](https://user-images.githubusercontent.com/26068101/34901558-6f474326-f7d1-11e7-935e-d8f6962ee03d.png)

shikhir-arora:
Does this happen with any song/URL? `.queue <some song>`  -- still the same error?

NeeAfk:
Yes it happens with song urls for youtube and just general searching. I tried a soundcloud link and it worked fine though, I should also mention I made a new api for the bot as well thinking that might have been the problem. Still doesn't work though @shikhir-arora .

shikhir-arora:
Can confirm it's not you and this is not working for YouTube in the current/latest version. I'll have to look further into it.

NeeAfk:
Glad at least someone else is having this problem. It sucks though cause it was **JUST** working...

shikhir-arora:
```bash

4|Nadeko   | 0 02:06:36 <GetResolveStrategy>d__2 | Getting link
4|Nadeko   | 0 02:06:37 <ResolveSong>d__2 | WARNING: unable to download video info webpage: HTTP Error 402: Payment Required
4|Nadeko   | WARNING: unable to download video info webpage: HTTP Error 402: Payment Required
4|Nadeko   | WARNING: unable to download video info webpage: HTTP Error 402: Payment Required
4|Nadeko   | WARNING: unable to download video info webpage: HTTP Error 402: Payment Required
4|Nadeko   | WARNING: unable to download video info webpage: HTTP Error 402: Payment Required
4|Nadeko   | Traceback (most recent call last):
4|Nadeko   |   File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
4|Nadeko   |     "__main__", fname, loader, pkg_name)
4|Nadeko   |   File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
4|Nadeko   |     exec code in run_globals
[...]

PR #1992, opened 2018-01-13
Fix for #1991

shikhir-arora:
Fixes https://github.com/Kwoth/NadekoBot/issues/1991 by forcing IPv4.


shikhir-arora:
^ And fix for that small warning with the latest changes ¯\_(ツ)_/¯

Kwoth:
what's -4

shikhir-arora:
Forces IPv4. One can also do `--source-address 0.0.0.0` to bind to all IP's, which does IPv4 by default obviously. But yeah, was able to replicate that and this fixed it.

PR #2067, opened 2018-02-08
Add double spaces (fixes formatting)

tukimoop:
hopefully.


Kwoth:
You should add it after every line? I mean examples too. It's kinda weird in this current format

tukimoop:
I've checked the previous placeholders file, it should be fine

PR #2080, opened 2018-02-11
AFK -> DoNotDisturb

shivaco:
There is no AFK status in `%bot.status%`. :^)


Macley-Kun:
Don't we call it DnD?

shivaco:
@Macley-Kun Could call it `DnD`, but since the bot outputs it as `DoNotDisturb`, I decided to put that unstead.

PR #2087, opened 2018-02-13
Fix the heck out of LeetCommand.

FurkanKambay:
I haven't seen anything like those switch cases in my life.
I tested with a few words, exact matches.

### DIFF
--- a/NadekoBot.Core/Modules/Games/LeetCommands.cs
+++ b/NadekoBot.Core/Modules/Games/LeetCommands.cs
@@ -6,10 +6,6 @@ using System.Collections.Generic;
 using System.Text;
 using System.Threading.Tasks;
 
-// taken from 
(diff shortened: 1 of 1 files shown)


FurkanKambay:
List of changes (Line numbers are for the new file):
(because the Diff looks so weird.)

- L18: There was an inconsistency between the other Games. So I created a class by properly with the attribute and base class.
- L23, L45: Little improvements.
- L43-75: Removed the switch-cases. Instead used a lookup strategy. The rest should be easy.
- L78: New lookup table.

Kwoth:
Woah, I forgot we had that there. It indeed is disgusting. Thanks so much. Just remove the top comment now as you rewrote it completely?

FurkanKambay:
No problemo!

Kwoth:
Lovely. Thanks a lot. Would love to see more of PRs like these haha.

PR #2088, opened 2018-02-13
Code Analysis & minor improvements

FurkanKambay:
Only noteworthy manual change is in these lines, it's an improvement:
https://github.com/Kwoth/NadekoBot/compare/1.9...FurkanKambay:1.9#diff-0764eabc1e78d5fdf150e5904a5304cfR111

Other than that, no functionality changes. Just `inline out var`s (IDE0018), spacing fixes and using statements sort&removal.


Kwoth:
👌

PR #2107, opened 2018-02-17
Changed Icon Image and Added Thumbnail Preview for Stream Sources

Kalimeme:
Fixed the Picarto User Icon and added a Stream Thumbnail Preview and User Icon for all of the Stream Sources.

klwuamf
ynkodec


Macley-Kun:
This looks amazing! There’s one thing I would suggest (not required),.

The url under the streamer name. Can that maybe be placed in the footer of the embed? (Idk I find the link and the streamers name a bit redundant ya know).

Kwoth:
I'll keep this in mind.

PR #2139, opened 2018-02-28
FIX: commands list designation

ingochris:
I simply fixed this to save a hop: https://nadekobot.me/commands

Best regards,
Christopher Ngo


Kwoth:
ty

Issue #2121, opened 2018-02-22
reactionroles not working with roles that are  more than one word


RapthorneLightweaver:
As title says, the .rero command seems to only work if you use a role name that is a single word only, is this intended?

Kwoth:
Yeah. It's not possible sorry.

FurkanKambay:
@Kwoth How about wrapping it with `"`s? Cmd does it. We can just say `argument.Trim('"')`

Macley-Kun:
@FurkanKambay is it a possibility that you test out this solution you have in mind, and maybe pull it over to nadekobot?

FurkanKambay:
@Macley-Kun just took a look at it. Discord allows characters like `"''` in the role name, so my suggestion won't work. But I think if we count every word before the :emoji: as a role name, that would work. I will try to do it when I find the time.

Macley-Kun:
Thanks dude! 👍

FurkanKambay:
@Macley-Kun #2142

PR #2142, opened 2018-03-02
.reactionroles improvement #2121

FurkanKambay:
Support for role names that are >1 words.
Counts every word before an :emoji: as part of the role name.
Untested; only in my head.

### DIFF
--- a/NadekoBot.Core/Modules/Administration/RoleCommands.cs
+++ b/NadekoBot.Core/Modules/Administration/RoleCommands.cs
@@ -1,4 +1,4 @@
-﻿using Discord;
+using Discord;
 using Discord.Commands;
 using Discord.WebSocket;
 using NadekoBot.Common;
@@ -8,6 +8,7 @@ using NadekoBot.Extensions;
 using NadekoBot.Modules.Administration.Services;
 using System;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 
 namespace NadekoBot.Modules.Administration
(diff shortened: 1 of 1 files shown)


Macley-Kun:
Jezus your fast! I'm sure kwoth will review the pull in a day or so. Thank for helping out!~<3

FurkanKambay:
@Macley-Kun You think that was fast? I wonder what you'd think if I was actually using an IDE or an editor or I didn't have to debug in my head. LOL, you're welcome <3

shikhir-arora:
Unfortunately this does not work and also breaks reaction roles. I tried the role name in quotes, as-is, with an emoji that was animated (which normally works), and also a normal request: `.rero Some Role :emoji:` -- and I did ensure that the emojis were in the guild. Does not work for any `.rero` type.

FurkanKambay:
@shikhir-arora Are you able to debug to see where it's failing?

FurkanKambay:
@shikhir-arora Oh, I think I see what's wrong. It's the regular expression. I'll try to fix it.

JohnGow383:
This already works, just use the .rid (role ID) instead of the words or encapsulate with quote marks

shikhir-arora:
Are you making sure to check the edge cases though? Remember `.rero` can take multiple roles:

`.rero Role :SomeEmoji: Role2 :SomeEmoji: Role3 :SomeEmoji:`

FurkanKambay:
@JohnGow383 How is that? The method  doesn't cover those cases as far as I can see. 

@shikhir-arora Yeah, I kept that in mind.

FurkanKambay:
Problem should be solved. I connected to my PC remotely and debugged, lol. The emoji string wasn't what I was expecting...

JohnGow383:
I have done this with roles with emojis, took a bit of time but for pokemon teams i have: 

.reactionroles excl "🔥 Valor" :val: "⚡ Instinct" :InstinctAni: "❄️ Mystic" :mys:

The role has the quotes @FurkanKambay

FurkanKambay:
@JohnGow383 Oh god. I wasted my time. I just thought since the method doesn't cover those cases, they wouldn't work, but it's done before the method call... 

Welp, TIL not to fix something before making sure it doesn't work.

shikhir-arora:
@FurkanKambay FWIW, your code after the fix regex did almost work 😆 - you just had to change it to something like this:

`var emojiRegex = @"<[a]?:?\w{2,32}:\d{17,19}>";` 

Mainly because you want to match animated avatars too (which are `<a:name:id>` vs `<name:id>`) and to clean it up the `name` has to be 30 or less characters while the digits have to be between 17-19, *technically*, since you could have some really old emoji's that have 17 count ID's.

But 👏

FurkanKambay:
@shikhir-arora Thanks! And thanks for the info, also TIL.

PR #2158, opened 2018-03-10
Add GlobalWhitelist/GlobalUnblocker Feature

Warped2713:
## Pull Request Status
- **Base Fork**: Kwoth/NadekoBot **Base**: 1.9
- **Head Fork**: BraveNetwork/LinBot **Compare**: gwl-nadeko-PR

- ✅  Currently up-to-date with NadekoBot/1.9 at commit e757e05
- ✅  Passed [(limited) test cases](https://docs.google.com/spreadsheets/d/15-S4Q22w0FL3Q25-7gBUZFUQnh7i8_fc4bJ58eIcOC4/edit?usp=sharing) at commit [ffff5f5](https://github.com/BraveNetwork/LinBot/commit/ffff5f51aa0357ef90d1a58479d0c3401a51ac1e) 
- ❌  Live preview not yet available
- ❌  Response String files still need Localization (or at least placeholders)
- ❌  The not OwnerOnly commands need to be moved to Utilities
- ❌  The priority on Add/Rm overloads needs adjustment

## About this Feature

Requested by NaviKing, Faeth, et al., the GlobalWhitelist/GlobalUnblocker feature allows bot owners to unblock modules or commands that have been blocked with `.gmod` or `.gcmd` for everyone, for specific users, channels, or servers, or for specific roles.

For example:

- globally disable **games**, and unblock just **pick**, **plant**, and **gencurrency** for everyone
- globally disable unrestricted commands, and unblock it for just the bot owners
- globally disable **music**, and unblock it for just one server
- globally disable your WIP features, and unblock it for Beta testers (role, server, channel, etc.)

You can also enable/disable a Global Whitelist at anytime in case you need to edit the members, or if you only want to unblock features on special occasions.

### Unblocking Rules

[A chart describing when a command/module is blocked or unblocked can be found here](https://docs.google.com/spreadsheets/d/1LznzrVqG0D0ym4L0iOEwKwIM89GsH0FaPNuVaa8bc2g/edit?usp=sharing)

- Global Permissions (gmod, gcmd) trump Local Permissions (sm, sc, etc.) when blocking a command or module.
- Local Permissions (sm, sc, etc.) trump anything Global Whitelist tries to unblock.
- Commands unblocked via Global Whitelist will work, regardless of whether the command, its module, or neither has been globally blocked (gmod, gcmd)
- Modules unblocked via Global Whitelist will allow all of their commands to work, regardless of whether the command, the module, or neither has been globally blocked (gmod, gcmd)

### Newly Added Commands

[A complete table of the newly added commands can be found here](https://docs.google.com/spreadsheets/d/1T-r8ZFdcsSGtqckqCj-Ob-EkjsBVk7vAJl2ksPE-rEc/edit?usp=sharing)

All commands are `OwnerOnly`, with the exception of the following:

- **mygwl** Get a list of all whitelists you belong to (both Member and Role -type lists)
- **ismygwl** Whether or not you belong to the given whitelist
- **gwlhere** Get a list of all whitelists for the current Server-Channel context (both Member and General -type lists)
- **isgwlhere** Whether or not the current Server or Channel belong to the given whitelist

- **mygub** Get a list of all commands and modules unblocked explicitly for you or your roles. This DOES NOT include disabled whitelists, nor does it include data from General-type whitelists.
- **ismygub** Whether or not the given command/module is unblocked for you or your roles. This checks BOTH disabled whitelists and General-type whitelists.
- **gubhere** Get a list of all commands and modules unblocked explicitly for the current Server or Channel. This DOES NOT include disabled whitelists, but it DOES include data from General-type whitelists.
- **isgubhere** Whether or not the given command/module is unblocked for the current Server or Channel context. This checks BOTH disabled whitelists and General-type whitelists.

#### Order of Arguments

In general, the order of arguments for any given GWL/GUB command is as follows:

1. WhitelistName `string`
2. FieldType `GEN` | `CMD` | `MDL` | `SRVR` | `CHNL` | `USR` | `ROLE` | `UB` | `MEM`
3. Any one of the following:
    1. ID `ulong`
    2. Name `string` | `Command` | `Module` | `Server` | `TextChannel` | `User` | `Role`
    3. Mention `#TextChannel` | `@User` | `@Role` (don't use FieldType for these!)
4. Enable/Disable `true` | `false` | `PermissionAction`
5. Page Number `int`

#### FieldTypes

FieldTypes help consolidate commands by requiring that the user specify what kind of data they are interested in.

When asking for whitelist info, you can use any FieldType to view only that type of data:

- `GEN` List all data for the whitelist
- `UB` List all unblocked commands and modules for the whitelist
- `CMD` List only unblocked commands for the whitelist
- `MDL` List only unblocked modules for the whitelist
- `MEM` List all servers, channels and users in the whitelist
- `SRVR` List only the servers in the whitelist
- `USR` List only the users in the whitelist
- `ROLE` List only the roles in the whitelist

When creating a whitelist, you specify what type of whitelist it should be with these FieldTypes: 

- `GEN` General whitelists unblock specific commands and modules for everyone everywhere
- `MEM` Member whitelists unblock specific commands and modules for specific servers, channels and users
- `ROLE` Role whitelists unblock specific commands and modules for specific Server-Role pairs

When adding/removing items from a whitelist, you can specify the type of data you are adding/removing with these FieldTypes: `CMD` | `MDL` | `SRVR` | `CHNL` | `USR` | `ROLE` 

FieldTypes when adding/removing are intended for use with lists of IDs or non-unique string names (such as when someone's username is also a module). If you use FieldTypes with a list of mentions, you risk letting the TypeReader think that the FieldType itself is a `#Channel`, `@User`, or `@Role`.

### Database Models

[Database Model](https://i.imgur.com/VMNjUvs.png)

### Checking a User's Roles

Since it is likely for a user to have many more roles than there are whitelisted roles, whenever we need to check if a command or module is unblocked for the user, we do the following:

1. Find all enabled Role whitelists for which the given command/module is unblocked (or just find all Role whitelists if not checking for a specific command/module)
2. Find all roles belonging to those whitelists
3. Count how many of those roles have this user in their membership

#### Logic

Since this will be checked every time a command is used, we take care not to check the database any more than necessary.

1. If not in a blocked hashset, let it pass.
2. If module is in blocked hashset, check if check if either the command or the module is in an unblocked hashset:
    1. If in unblocked hashset, 
        1. If either is in any enabled General whitelists, let it pass.
        2. Otherwise if either is in any enabled Member whitelists, let it pass.
        3. Otherwise if either is in any enabled Role whitelists, let it pass.
        4. Otherwise report the blocked module.
    2. If not in an unblocked hashset, report the blocked module.
3. If command is in blocked hashset, check if either the command or the module is in an unblocked hashset:
    1. If in unblocked hashset, 
        1. If either is in any enabled General whitelists, let it pass.
        2. Otherwise if either is in any enabled Member whitelists, let it pass.
        3. Otherwise if either is in any enabled Role whitelists, let it pass.
        4. Otherwise report the blocked command.
    2. If not in an unblocked hashset, report the blocked command.
4. Otherwise report blocked command or module.

#### Reporting

As requested by NaviKing, the `GlobalPermissionService.TryBlockLate()` will report that a command/module has been blocked, similarly to what `.verbose` does for Local Permissions.

To avoid message spam, both the report message and the offending message will be deleted after 10 seconds.

### Long Command Usage Text

Given the many method overloads and FieldTypes allowed by the GWL/GUB commands, the command usage text gets very long. For this reason, `NadekoBot.Core/_Extensions/Extensions.cs` has been edited such that if there is more than 2 provided usages, they get formatted as one usage per line in a single code-block.

#### Example: 2 or Fewer
`.usage 1` or `.usage 2`

#### Example: More than 2

```
.usage 1
.usage 2
.usage 3
```


Kwoth:
While impressive, this is just too huge to be accepted as it will require maintenance from my side, and it's not how i'd do it in the first place. 

Instead of doing this, you could've just abstracted out the permission system checking logic and basically create a parallel permission system which will be bot owner and global, only and work on server/channel/user basis  instead of server/channel/role/user.

However I'd like it to stay open so that people who are modifying their bots can add it to their local versions if that's fine.

Warped2713:
The reason I included support for roles is because my design is for many shallow lists. For example, a list of 100 users could be flattened into a list of maybe 2 roles or 5 servers, which makes it easier to maintain on the bot owner's end and keeps the number of database records low.

I'm not totally sure I understand what your version of this would look like, but since this was my first time using Discord.NET and EF Core, I made this as a sort of combination of the existing Blacklist and GlobalPermissions designs. Plus, coming from a MySQL background, my instinct was to use Many-To-Many relationships in the DB.

But if you're worried about maintenance, I don't mind being the go-to person for keeping this up-to-date with Nadeko since I will have to do this anyway for LinBot.

I'm fine with this staying open, if you still don't want to add it officially.

TiBarification:
@Warped2713 , if you will try to add a role, that is higher than bot role, it prints that message
default
Type of whitelist is `ROLE`

shikhir-arora:
You aren't supposed to be able to do that, it's a Discord limitation. If this is using that then it won't work. It says `@deleted-role` probably because the role has a space in it (known thing), but that's not the issue.

If you want to let the bot get a role higher than it, and assign it to a whitelist, you'd have to do it outside of D.Net - so you'd have to do basically what Kwoth stated: abstract out the permissions system, unrelated from Discord perms.


tl;dr if you use D.Net, you're not going to be able to assign a higher role and then use that successfully. This is likely why you're not getting the access; adding to the whitelist is done but it won't actually work because the bot isn't able to do that.

Warped2713:
@TiBarification I don't believe the problem is "role higher than bot's role" as I can add Admin roles just fine.

Did you make sure to specify the correct Server for your role? @deleted-role appears when the Discord app tries to resolve a mention of a role that doesn't exist in the current server.

Warped2713:
@TiBarification Another possibility is that if you tried to copy the Role ID by rightclicking a message, you used the message ID and not the Role ID. Use `\@Role` to get the true Role ID

TiBarification:
>Another possibility is that if you tried to copy the Role ID by rightclicking a message, you used the message ID and not the Role ID. Use \@Role to get the true Role ID
Thanks, it works. I was acidentally added here `message id` instead of role id.

TiBarification:
Just in case it would be nice to add check if that a role, not a channel 👍

Kwoth:
I'll close this now

Issue #1832, opened 2017-11-15
.xp doesn't show unicode characters on image


TiBarification:
How it works:
![default](https://user-images.githubusercontent.com/6222472/32847744-882de88e-ca33-11e7-9c90-24605205b331.png)

Kwoth:
This is because the drawing library I'm using doesn't support drawing unicode characters :c

Macley-Kun:
Should we make an issue on the library your using instead? Because even with help this is out of your doing :P

coconut24:
the problem is within the font, the drawing library will draw nothing if the font doesnt support the characters

Macley-Kun:
Tested this out on NadekoBot v2.10.1. Still present.

Macley-Kun:
Kwoth, can you make an issue on the drawing library and ask if they wanna add unicode characters?

Macley-Kun:
Update:

Rjt suggested to use Noto Sans. This font supports more characters what should solve this issue.

florian-kaiser:
What about just using the Arial font? I tested it and it draws the unicode character perfectly

Kwoth:
Nice try, but it doesn't work properly, and having the single font be huge % of the size of the whole bot is not viable. It works sometimes but most of the time not. It's better than what we currently have for sure, but definitely not worth it.

![image](https://user-images.githubusercontent.com/2537696/37268142-0f4a6b98-25f7-11e8-9102-0e865f15607f.png)

Kwoth:
Tried it, it's not the font, it's the fact that the library has issues drawing them I believe.

kronoize:
What library is it drawing from? Because you can always just download Microsoft IMEs for the languages most commonly used in the usernames like JP and minimize storage or download all of them with more compatibility with usernames at the cost of more storage.

Though if you're already using it and it's still having problems with drawing from it, just ignore this message I guess.

zMatix:
How to change background like you?

PR #2161, opened 2018-03-11
adding unicode arial font

cywq:
Added Arial Unicode MS font for the .xp command and general use.

font size is 22mb but contains the most needed unicode glyphs.

Maybe we can update to unicode 10.0 sometime.
199179994520354816_xp


TiBarification:
closes #1832

Kwoth:
#1832 read my reply, it doesn't. If someone has different results please share.

Issue #2162, opened 2018-03-12
gambling_start_event_fail key is missing from en-US response strings. When creating an even with any parameters


Coldvictory:
`0 01:38:22 CurrencyEventsService | Discord.Net.HttpException: The server responded with error 400: BadRequest
   at Discord.Net.Queue.RequestBucket.<SendAsync>d__18.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Discord.Net.Queue.RequestQueue.<SendAsync>d__14.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Discord.API.DiscordRestApiClient.<SendInternalAsync>d__62.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Discord.API.DiscordRestApiClient.<SendAsync>d__51.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Discord.API.DiscordRestApiClient.<AddReactionAsync>d__84.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Discord.Rest.MessageHelper.<AddReactionAsync>d__2.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #2165, opened 2018-03-12
Fixes missing gambling_start_event_fail key

cywq:
fixed a misspelling in the ResponseStrings US #


cywq:
fixes #2162

Kwoth:
Oh thanks. I'll just make this change manually as I've made a bunch of changes for the upcoming 2.12 so this would create an unnecessary conflict

PR #2278, opened 2018-05-05
Add count replacement ("searches_streams_cleared")

zvjake:
POEditor string:
"All {0} followed streams on this server have been removed."
Function stores count var.
All makes it seem like the bot should be using count as a replacement.


Kwoth:
Nice catch, thank you.

PR #2281, opened 2018-05-06
Add OkColor to .lsar

Gremagol:
Add the OkColor back to .lsar because after the Update the Color is gone.
Before:
lsarbefore
And after:
lsarafter

Also yay my first PR 😂


Kwoth:
Thank you

PR #2342, opened 2018-06-04
Handle non-200 HTTP responses in .randomcat and .randomdog commands.

### LINKED ISSUE #2336: .meow command is not working
Bot hosted on Amazon AWS, OS is Ubuntu 16.04.4 LTS
Nadeko version is 2.22.0

This is what I get in the console.
meow
.woof command works just fine.

FurkanKambay:
Closes #2336 

I'm not very familiar with the codebase; can improvements like this be made elsewhere?

If there is a helper method to check for a non-`200` status code, we'll make use of that. If there isn't, we can write one (e.g. a method that returns a `bool` success and the content as `out string`) I presume that could be useful in other places.


Kwoth:
Randomcat and randomdog have already been reworked locally, but yes, elsewhere it might be possible to give nicer errors to users on commands that fail.

PR #2372, opened 2018-06-23
Prevent Discord from caching stream previews

zfbx:
Added a random number version string to the end of the preview URL to prevent Discord from caching the preview which was causing every stream to use the same preview image. *could also be done with something like a timestamp but I thought this would come out cleaner.


Kwoth:
Ah, thanks. I will look into that.

PR #2394, opened 2018-07-05
add utf-8 support for non-english langs google searchs

EduenSarceno:
added query params `ie` & `oe` to force utf-8 treatment, also embed title changed
it will print the original search terms (this avoid the annoying percent-encoding
related to non-english langs)

before vs after


Kwoth:
can you please fix image search too?

Kwoth:
I've squashed the commits and pushed them myself with credits to you.

PR #2424, opened 2018-07-16
Enforce Bold Font Style for Noto Sans

rjt-rockx:
Prevents the Regular Font Style being selected when Noto Sans fonts are previously installed.


Kwoth:
ty

PR #2426, opened 2018-07-17
Allow for 0 values of TriviaCurrencyReward

theLAZYmd:
Default value is 0 anyway.
Should be allowed to set it back to 0 if decided the feature is no longer wanted.
At the moment, tryParse && triviaReward > 0 is blocking that


Kwoth:
thank you

PR #2457, opened 2018-07-26
Added Translation for ResponseStrings.zh-TW

jin940:
Translation from English to Chinese(Taiwan).


Kwoth:
Thanks, but this is not how strings should be updated. You can DM me on discord in order to add you to POeditor chinese translator group.

PR #2462, opened 2018-07-28
fixed .crypto, fixes #2458

### LINKED ISSUE #2458: Crypto command not pulling correctly
I'm into an ERC-20 utility token called Refereum (RFR) and I'm a big help in their community. As of late, I've noticed that there is a Crypto command in Nadeko. When attempting to use it to track stats of cryptos, I notice I can pull all kinds and that it uses coinmarketcap to get the info. However when attempting to check Refereum, it states "Cryptocurrency with that name was not found." I find this odd if it's pulling from coinmarketcap since Refereum is listed there. https://coinmarketcap.com/currencies/refereum/#markets
Is there a way to add cryptos to this command? Or can it possibly be updated in a future patch/version to include this crypto? It's been around for nearly a year now and I still can't get the command to work for it.

ia03:


Kwoth:
Oh I've already fixed crypto, just didn't push yet.  I've moved crypto stuff to its own service and made it use new api. the guy who made the issue got flagged for some reason so his issue went away :P

PR #2463, opened 2018-07-28
Updated Response Strings (pt-BR / July 2018)

Kaoticz:
This version contains massive general improvements across 48% of all strings. I hadn't commited it  before because ~~I didn't know I had to~~ I'm dumb. Sorry for the inconvenience.


Macley-Kun:
I think it’s would be better to do these change in Poe editor. Ask Kwoth if you can join the program.

Kaoticz:
This is the file generated by POEditor. I didn't touch it :p

Macley-Kun:
Oh ok then should be good :-) thx for translating!

Outlook voor iOS<https://aka.ms/o0ukef> downloaden
________________________________

PR #2474, opened 2018-08-02
Added a simple .invite command.

Gremagol:
Hi, I've made a simple .invite command, which uses the given ClientID from the Crendentials, so selfhosts should work too.

I also added the command string and output string to the JSON's. The translators can use the `{0}` for the Invite link. 

It looks like this right now:
image

If it's unneeded or useless, just close this PR. 😃 
Have a nice day!


Kwoth:
in the next patch `.acr .invite ....invitelink...&client_id=%bot.id%` will work for most bots (except really old ones where client id is not the same a bot id)

PR #2517, opened 2018-08-20
Adding shop management commands and Raspbian dependency

casmbu:
Added shopswitch, shopprice, and shopname. Also added SQLitePCLRaw.lib.e_sqlite3.linux to NadekoBot.Core which enables this to be built for Raspbian.


Kwoth:
Thanks, but the shop system will get a full rewrite, it's not intuitive,.

Issue #2533, opened 2018-08-24
FEATURE: configure the location of redis server


LordVeovis:
Hi,

I'm trying to write a Dockerfile to have a more up-to-date version of Nadeko that what is available from willysunny. At least one by tag.
This new container will be lighter, simpler and only contain the nadeko process, leaving redis for another container, which respect one of the container principles: "1 container = 1 service".

Let's go to the fact. During my test, I've noticed that the redis server is hardcoded in the source code : https://github.com/Kwoth/NadekoBot/blob/1.9/NadekoBot.Core/Services/ShardsCoordinator.cs L81

I think this setting (hostname and optionally also the port) MUST be configurable.
Maybe on the credentials.json file or as an environnent variable (easier to implement ^^) ?

What do you think?

Macley-Kun:
What about this docker? https://github.com/shikhir-arora/Nadecker

Auto updates, easy install (even easier then Linux (aside of installing docker itself)).

Outlook voor iOS<https://aka.ms/o0ukef> downloaden
________________________________

LordVeovis:
Thanks for the suggestion but that's still wrong for the other points:
* a container should NOT auto-update itself. The purpose of a tagged container is to exactly know what version is running. An update MUST imply changing the container. In other words: a healthy container is immutable
* redis is still in the same container than nadeko, preventing a good monitoring of the nadeko process (hidden behing the auto-restart.sh). Anyway the auto-restart feature should be controlled by dockerd itself
* well... debian/ubuntu base is klunky at best for a container: either fat or slim with no basic command tool to debug. Currently my image is only 228MB and with a complete linux tools, a quarter of willysunny's.

LordVeovis:
I've been able to starts the container with only Nadeko inside: the bot was able to respond to my .h on Discord.
I had to patch 127.0.0.1 2 times with another hardcoded value (just for testing).
I also had to patch the default ShardRunArguments as the program did not take the one defined in the credentials.json, I don't know why.

```
sed -i 's/127.0.0.1/redis/' NadekoBot/NadekoBot.Core/Services/ShardsCoordinator.cs; \
sed -i 's/127.0.0.1/redis/' NadekoBot/NadekoBot.Core/Services/Impl/RedisCache.cs; \
# for linux, I want dotnet my.dll {0} {1}
[...]

PR #2536, opened 2018-08-25
Added Redis Options #2599

LordVeovis:
Hi @Kwoth,

This achieves the modifications on the Redis options, related to the issue #2533 
Tested: the bot connects successfully to Discord.

### DIFF
--- a/NadekoBot.Core/Services/Impl/BotCredentials.cs
+++ b/NadekoBot.Core/Services/Impl/BotCredentials.cs
@@ -80,6 +80,7 @@ namespace NadekoBot.Core.Services.Impl
                 CleverbotApiKey = data[nameof(CleverbotApiKey)];
                 MiningProxyUrl = data[nameof(MiningProxyUrl)];
                 MiningProxyCreds = data[nameof(MiningProxyCreds)];
+                RedisOptions = data[nameof(RedisOptions)];
 
                 VotesToken = data[nameof(VotesToken)];
                 VotesUrl = data[nameof(VotesUrl)];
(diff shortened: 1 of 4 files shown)


NadekoBot.Core/Services/Impl/BotCredentials.cs
```diff
@@ -44,7 +44,7 @@ public class BotCredentials : IBotCredentials
         public string VotesUrl { get; }
         public string VotesToken { get; }
         public string BotListToken { get; }
-        public string RedisOptions { get; }
+        public string RedisOptions { get; } = "127.0.0.1,syncTimeout=3000";
```
Kwoth: instead of putting this here, you should put it in the else statement under line 83, since like this it will be in people's credentials always and if i decide to change something with it in the future for everyone who hasn't made a custom setting i won't be able to.

docs/JSON Explanations.md
```diff
@@ -25,6 +25,7 @@ If you do not see `credentials.json` you will need to rename `credentials_exampl
   "ShardRunArguments": "",
   "ShardRunPort": null,
   "TwitchClientId": null,
+  "RedisOptions": "127.0.0.1,name=nadeko"
```
Kwoth: this should be null or an empty string.

NadekoBot.Core/Services/Impl/BotCredentials.cs
LordVeovis: ok for that but I may miss something as I don't see the functional difference between:
* pre-initialize RedisOptions with a default value and override it if a value is present in the credentials.json file (current implementation)
* populating RedisOptions from credentials.json if any or take the default value (if/else proposition)

Can you elaborate a bit more ?

Kwoth: Oh, yea. I thought you put the value in the other model which gets serialized into credentials_example.json, guess it's just more consistent like this at least :P

Kwoth:
Thanks!

Issue #2544, opened 2018-08-29
Embed footer's icon_url doesn't parse placeholders


zergling-man:
https://github.com/Kwoth/NadekoBot/blob/1.9/NadekoBot.Core/Common/Replacements/Replacer.cs#L59

PR #2546, opened 2018-08-29
Embed footer's icon_url fix?

Gremagol:
Possible fix for issue #2544 
Before:
image
And after:
image


Gremagol:
I'll close it since it seems like you fixed it already.

PR #2551, opened 2018-09-01
Fix for servers with a lot of long emoji names

Anheledir:
This caused the following exception on one of my servers:
System.ArgumentException: Field value length must be less than or
equal to 1024.


Kwoth:
Thank you

PR #2565, opened 2018-09-13
Fix youtube videoid

LordVeovis:
Some Youtube video's id starts with a minus.
Given to youtube-dl without escaping, it handles the video id as a switch.

Sample: https://www.youtube.com/watch?v=-5DyDpdBW3A

I had to add a debug line, that in my sense should not be removed to better understand stack traces but be controlled by the nlog conf (PR incoming).


Kwoth:
thanks!

Issue #2556, opened 2018-09-05
FEATURE: auto-build a docker image


LordVeovis:
Objective of this new feature: propose an automatically built Docker image

Benefits:
* very simple (aka easy to review or maintain) build script
* light (280MB vs 2GB with the current available images)
* simple architecture (1 docker = 1 service)
* auto-build a latest tagged image when something is pushed on branch 1.9
* auto-build a version tagged image when a new git tag is created (or some tags only by using a regexp)

The two last points are only available when the Dockerfile is part of the source code when using the autobuild feature of the dockerhub platform.

This is the using Dockerfile I'm using currently. This Dockerfile is not the one to be merged, it has an extra step which is the git clone (step already done by the CI tool usually)
```
FROM microsoft/dotnet:2.1-sdk-alpine AS build

ENV TAG=1.9

RUN set -ex; \
    apk add --no-cache git; \
    git clone -b "$TAG" https://github.com/LordVeovis/NadekoBot

WORKDIR /NadekoBot/src/NadekoBot
RUN set -ex; \
    dotnet restore; \
    dotnet build -c Release; \
    dotnet publish -c Release -o /app

WORKDIR /app
RUN set -ex; \
    rm libopus.so libsodium.dll libsodium.so opus.dll; \
    find . -type f -exec chmod -x {} \;; \
    rm -R runtimes/win* runtimes/osx* runtimes/linux-*

FROM microsoft/dotnet:2.1-runtime-alpine AS runtime
WORKDIR /app
COPY --from=build /app /app
RUN set -ex;\
    apk add --no-cache \
        ffmpeg \
        youtube-dl \
        libsodium \
        opus; \
    adduser -D nadeko; \
    chown nadeko /app; \
    chmod u+w /app; \
    install -d -o nadeko -g nadeko -m 755 /app/data

# workaround for the runtime to find the native libs loaded through DllImport
RUN set -ex; \
    ln -s /usr/lib/libopus.so.0 /app/libopus.so; \
    ln -s /usr/lib/libsodium.so.23 /app/libsodium.so

VOLUME [ "/app/data" ]
USER nadeko
CMD ["dotnet", "/app/NadekoBot.dll"]
```

And the docker-compose.yml to run the Nadeko stack:
```
version: "3.5"
services:
  nadeko:
    #image: willysunny/nadecker:latest
    build: .
    depends_on:
      - redis
    restart: unless-stopped
    environment:
      TZ: Europe/Paris
    volumes:
[...]

PR #2566, opened 2018-09-13
New docker

LordVeovis:
Let's starts the implementation of #2556

This PR uses a more recent version of youtube-dl, the very-old one was not reliable depending on the youtube playlists.

The documentation is... laking... but I'll try my best to have a readable noob-proof guide if desired.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,39 @@
-FROM microsoft/dotnet:sdk
-MAINTAINER Poag <poag@gany.net>
+FROM microsoft/dotnet:2.1-sdk-alpine AS build
 
-WORKDIR /opt/
+COPY . /nadekoBot
 
-#Install required software
-RUN echo "deb http://www.deb-multimedia.org jessie main non-free" | tee /etc/apt/sources.list.d/debian-backports.list \
-	&& apt-get update \
-	&& apt-get install -y --force-yes deb-multimedia-keyring \
-	&& apt-get update \
-	&& apt-get install -y git libopus0 opus-tools libopus-dev libsodium-dev ffmpeg
+WORKDIR /nadekoBot/src/NadekoBot
+RUN set -ex; \
+    dotnet restore; \
+    dotnet build -c Release; \
+    dotnet publish -c Release -o /app
 
-#Download and install stable version of Nadeko
-RUN curl -L https://github.com/Kwoth/NadekoBot-BashScript/raw/master/nadeko_installer_latest.sh | sh \
-	&& curl -L https://github.com/Kwoth/NadekoBot-BashScript/raw/master/nadeko_autorestart.sh > nadeko.sh \
-	&& chmod 755 nadeko.sh
+WORKDIR /app
+RUN set -ex; \
+    rm libopus.so libsodium.dll libsodium.so opus.dll; \
+    find . -type f -exec chmod -x {} \;; \
+    rm -R runtimes/win* runtimes/osx* runtimes/linux-*
 
-VOLUME ["/opt"]
+FROM microsoft/dotnet:2.1-runtime-alpine AS runtime
+WORKDIR /app
+COPY --from=build /app /app
+RUN set -ex; \
+    echo '@edge http://dl-cdn.alpinelinux.org/alpine/edge/community' >> /etc/apk/repositories; \
+    apk add --no-cache \
+        ffmpeg \
+        youtube-dl@edge \
+        libsodium \
+        opus; \
+    adduser -D nadeko; \
+    chown nadeko /app; \
+    chmod u+w /app; \
+    install -d -o nadeko -g nadeko -m 755 /app/data
 
-CMD ["/opt/nadeko.sh"]
+# workaround for the runtime to find the native libs loaded through DllImport
+RUN set -ex; \
+    ln -s /usr/lib/libopus.so.0 /app/libopus.so; \
+    ln -s /usr/lib/libsodium.so.23 /app/libsodium.so
+
+VOLUME [ "/app/data" ]
+USER nadeko
+CMD ["dotnet", "/app/NadekoBot.dll"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,15 @@
+version: "3.5"
+services:
+  nadeko:
+    build: .
+    depends_on:
+      - redis
+    volumes:
+      - ./credentials.json:/app/credentials.json:ro
+      - ./data:/app/data
+  redis:
+    image: redis:alpine
+    ports:
+      - 6380:6379
+    sysctls:
+      net.core.somaxconn: 511
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,5 @@ services:
       - ./data:/app/data
   redis:
     image: redis:alpine
-    ports:
-      - 6380:6379
     sysctls:
       net.core.somaxconn: 511
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,8 @@ version: "3.5"
 services:
   nadeko:
     build: .
+    #image: kwoth/nadeko
+    restart: unless-stopped
     depends_on:
       - redis
     volumes:
@@ -11,3 +13,4 @@ services:
     image: redis:alpine
     sysctls:
       net.core.somaxconn: 511
+
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,8 +9,14 @@ services:
     volumes:
       - ./credentials.json:/app/credentials.json:ro
       - ./data:/app/data
+    labels:
+      "com.centurylinklabs.watchtower.enable": "true"
   redis:
     image: redis:alpine
     sysctls:
       net.core.somaxconn: 511
-
+  #watchtower:
+  #  image: v2tec/watchtower
+  #  volumes:
+  #    - /var/run/docker.sock:/var/run/docker.sock
+  #  command: --interval 86400 --cleanup --label-enable
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+.vscode
+.vs
+data
+credentials.json
+_libs
+*.dll
+*.so
--- a/docs/guides/Docker Guide - Legacy.md
+++ b/docs/guides/Docker Guide - Legacy.md
@@ -1,0 +1,112 @@
+# Setting up NadekoBot on Docker
+Nadeko is written in C# and Discord.Net for more information visit <https://github.com/Kwoth/NadekoBot>
+
+## Before you start ...
+
+... If your PC falls under any of the following cases, please grab Docker Toolbox instead.
+
+For Windows [[Download Link](https://download.docker.com/win/stable/DockerToolbox.exe)]
+- Any Windows version without Hyper-V Support
+- Windows 10 Home Edition
+- Windows 8 and earlier
+
+For Mac [[Download Link](https://download.docker.com/mac/stable/DockerToolbox.pkg)]
+- Any version between 10.8 “Mountain Lion” and 10.10.2 "Yosemite"
+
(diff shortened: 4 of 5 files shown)


Kwoth:
Yea, we definitely need a guide to set it up. You can make it as the "preferred way" in the docs, and have the current guide be an alternative.

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+    build: .
+    depends_on:
+      - redis
+    volumes:
+      - ./credentials.json:/app/credentials.json:ro
+      - ./data:/app/data
+  redis:
+    image: redis:alpine
+    ports:
+      - 6380:6379
```
chauffer: any reason to expose ports by default? I assume most people would not need to access redis in any way other than through the bot

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+    sysctls:
+      net.core.somaxconn: 511
```
chauffer: is this necessary? my redis containers always work fine with no such changes

Dockerfile
```diff
@@ -1,20 +1,39 @@
-FROM microsoft/dotnet:sdk
-MAINTAINER Poag <poag@gany.net>
+FROM microsoft/dotnet:2.1-sdk-alpine AS build
 
-WORKDIR /opt/
+COPY . /nadekoBot
```
shikhir-arora: This should be `/NadekoBot` for consistency.

Dockerfile
```diff
@@ -1,20 +1,39 @@
...
 
-#Install required software
-RUN echo "deb http://www.deb-multimedia.org jessie main non-free" | tee /etc/apt/sources.list.d/debian-backports.list \
-	&& apt-get update \
-	&& apt-get install -y --force-yes deb-multimedia-keyring \
-	&& apt-get update \
-	&& apt-get install -y git libopus0 opus-tools libopus-dev libsodium-dev ffmpeg
+WORKDIR /nadekoBot/src/NadekoBot
```
shikhir-arora: See above; should be `/NadekoBot/src/NadekoBot`

docker-compose.yml
shikhir-arora: Yeah, the somaxconn is a good thing to enforce. Looks good 👍

Dockerfile
LordVeovis: The use of a caps on a case-sensitive filesystem as the first letter of a folder/file is just so annoying.

I don't see any consistency need (and compared to what) for the folder name containing Nadeko stuff. The container does not expose this folder.

Moreover if the bot is well-coded and the container is well designed, this should have no impact.

shikhir-arora: It may not, but it looks off - considering the native file structure is like so, and for general conventions I'd think going with that or just all lower case is fine, but it's not very normal to see `lowerUpper` in many files.

LordVeovis: Well, I'm also lazy so I'm just typing nad then press the TAB key. As long as the first letter is not a caps, I'm fine.

LordVeovis:
It depends on what you have put in your docker-compose.yml (or docker run equivalent).
By default, the bot die but docker can restart it depending on how the bot died with the restart parameter (https://docs.docker.com/compose/compose-file/#restart)
I find this very useful to control the restart of my services the same way. It also have the advantage to display the uptime of the process when you do a docker ps (aka the last restart).

shikhir-arora:
> It depends on what you have put in your `docker-compose.yml` (or `docker run` equivalent).
> By default, the bot die but the docker engine can restart it depending on how the PID=1 process died with the restart parameter (https://docs.docker.com/compose/compose-file/#restart)
> I find this very useful to control the restart of my services the same way. It also have the advantage to display the uptime of the process when you do a `docker ps` (aka the last restart).

You can also just make sure to specify a `RestartCommand` (default null) in the credentials, since `.die` isn't tied to Docker's operation. I don't think restart using Docker would work here (unless you manually restart the container), as the container likely isn't registered as dead with the die command.

Dockerfile
shikhir-arora: I think keeping it at `/nadekobot/` is more ideal for sure.

LordVeovis:
The `RestartCommand` is particular as it is also used to start a shard, with a minimal shard of 1, we can't null it yet.
Below is the process tree of my bot with one shard:
```
/app $ ps f
PID   USER     TIME   COMMAND
    1 nadeko     4:02 dotnet /app/NadekoBot.dll
   27 nadeko    87:49 dotnet /app/NadekoBot.dll 0 1
  619 nadeko     0:00 ffmpeg -reconnect 1 -reconnect_streamed 1 -reconnect_dela
```

Dockerfile
LordVeovis: I want to amend the /nadekobot. I was not watching the right stage. This Dockerfile is multi-staged and only the last one is the final one.
Actually, everything is in /app and /app is the WORKDIR:
```
sv-alp-docker-1:/docker/nadeko# docker exec -it -u root nadeko_nadeko_1 sh
/app # ls -lh /app/
total 17270
-rw-r--r--    1 root     root        4.2M Sep 10 20:52 NadekoBot.Core.dll
-rw-r--r--    1 root     root         529 Sep  5 20:37 credentials.json
-rw-r--r--    1 root     root         667 Sep 10 20:52 credentials_example.json
drwxr-xr-x    6 nadeko   nadeko      3.4K Sep 16 19:51 data
lrwxrwxrwx    1 root     root          21 Sep 10 20:53 libopus.so -> /usr/lib/libopus.so.0
lrwxrwxrwx    1 root     root          24 Sep 10 20:53 libsodium.so -> /usr/lib/libsodium.so.23
drwxr-xr-x    3 root     root        3.4K Sep 10 20:52 [a lots of files...]
```

shikhir-arora:
It's null by default. RestartCommand isn't the same as die, sorry if that wasn't clear. It lets you use `.restart` If you've configured something, otherwise it doesn't do anything.

shikhir-arora:
I would suggest adding an optional watchtower to the compose file. This should all be documented though. For example, we have something like:

```
watchtower:
    container_name: watchtower
    hostname: watchtower
    restart: always
    image: v2tec/watchtower
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    command: --interval 300 --cleanup
```

For those that want to pull automatically when there's an update. This one would be universal to all the Docker apps, but can be changed.

docker-compose.yml
chauffer: where is it mentioned? I cannot find it here https://hub.docker.com/_/redis/, and even at my workplace we run redis at a pretty significant scale without the need of this.

shikhir-arora: You can read about it here, for one: https://www.techandme.se/performance-tips-for-redis-cache-server/

Or just google it and you'll see tons of docker posts, but it's very common that many servers have the max TCP backlog connection parameter set as something lower than the minimum recommended.

```
WARNING: The TCP backlog setting of 511 cannot be enforced because /proc/sys/net/core/somaxconn is set to the lower value of 128
```

So you'll see something like the above on redis startup. It usually won't make a difference as there's unlikely to be that many connections in the backlog anyways, but it's a simple / harmless thing to increase unless you have a specific case not to. You can do it with sysctl on Linux so that's all it's doing (since the Redis image is based on Alpine)

LordVeovis: @chauffer I don't want to be rude, but that's the first thing displayed when you start a redis node.
```
/ # sv-alp-docker-1:/docker/mattermost# docker run --rm -it redis:alpine
1:C 17 Sep 12:41:43.942 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
1:C 17 Sep 12:41:43.942 # Redis version=4.0.10, bits=64, commit=00000000, modified=0, pid=1, just started
1:C 17 Sep 12:41:43.942 # Warning: no config file specified, using the default config. In order to specify a config file use redis-server /path/to/redis.conf
                _._
           _.-``__ ''-._
      _.-``    `.  `_.  ''-._           Redis 4.0.10 (00000000/0) 64 bit
  .-`` .-```.  ```\/    _.,_ ''-._
 (    '      ,       .-`  | `,    )     Running in standalone mode
 |`-._`-...-` __...-.``-._|'` _.-'|     Port: 6379
 |    `-._   `._    /     _.-'    |     PID: 1
  `-._    `-._  `-./  _.-'    _.-'
 |`-._`-._    `-.__.-'    _.-'_.-'|
 |    `-._`-._        _.-'_.-'    |           http://redis.io
  `-._    `-._`-.__.-'_.-'    _.-'
 |`-._`-._    `-.__.-'    _.-'_.-'|
 |    `-._`-._        _.-'_.-'    |
  `-._    `-._`-.__.-'_.-'    _.-'
      `-._    `-.__.-'    _.-'
          `-._        _.-'
              `-.__.-'

1:M 17 Sep 12:41:43.944 # WARNING: The TCP backlog setting of 511 cannot be enforced because /proc/sys/net/core/somaxconn is set to the lower value of 128.
1:M 17 Sep 12:41:43.944 # Server initialized
1:M 17 Sep 12:41:43.944 # WARNING overcommit_memory is set to 0! Background save may fail under low memory condition. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
1:M 17 Sep 12:41:43.944 * Ready to accept connections
```

chauffer: You're not being rude at all. :) However, I think simplicity is a better approach here. The non-obvious downside to this is it's not trivial to what's this doing, and it is simply not needed. https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

LordVeovis: YAGNI is not relevant here. I'm not developing anything here, just configuring. 
Futhermore as mentioned in the link, this alone is the best way to gain free technical debt. In other words, YAGNI is a principle to help you organize what features developing and how far in an agile way (from a MVP to another one), not to ship a half-done product because "apparently it seems to work". It should not deviate you from the state of the art as a final goal, which we have here.

LordVeovis:
@shikhir-arora ok on RestartCommand. I was confused with ShardRunComand.

About wathtower, I strongly suggest a big warning when using it. As awesome as it may be, this has a direct control on dockerd without any limitations. I'm not very please with this kind ot containers.
But this should please enough users to be documented in the doc, I'm OK with that.

shikhir-arora:
The bit I posted does, but it doesn't have to bind to the master docker sock if you really want it to only monitor a specific application. We do this for some other things. Anyways, most people would likely find watchtower useful globally and most probably won't run much docker outside of this. I run quite a few and two Kubernetes clusters on GCE, so I wouldn't consider it indicative of most users. The pic below is just one server's Docker containers. That's not a bad thing as most HA apps should be using Docker or orchestration 🙂

Looks good to be though, just make sure the docs are clear about linking credentials, installing compose if you include it (which is how I really only use Docker) - and compose makes that easier for you to document since it's just one command. 

Appreciate the time answering all my comments, so looks good to me on the technical side. Just the docs and @Kwoth would probably chime in and go from there. Thanks for the PR ✅ 


94d11962-5943-4f95-b274-663abbfb9d2b

docker-compose.yml
chauffer: I don't think I am going to be able to change your mind, but I disagree with you. Both will work, but this is unneeded lines which are not simple to understand, and not needed for correct functioning.

Kwoth:
What's the status on this? 

I see the docs are not yet updated

LordVeovis:
Sorry for the delay, I hope the doc is clear enough. 

@Kwoth I see you already own kwoth/nadeko on dockerhub. You will just have to customize the build settings for this with something similar to:

image

- for the kwoth/nadeko:latest be the unstable bleeding-edge version
- for the kwoth/nadeko:2.28.4 be the version corresponding to the git tag 2.28.4
And all that, automatically :)

PR #2597, opened 2018-10-01
.economy ru-RU strings

Jengas:
I have translated .economy strings


Kwoth:
Strings are updated on poeditor, not directly in this repo. We already have russian translators working on the translation. Thanks anyway.

PR #2615, opened 2018-10-12
Added `--safe` parameter to .prune command to skip pinned messages

Anheledir:


Kwoth:
Oh, cool. Can you just change `-safe` to `--safe`. Thanks

Kwoth:
Thanks a lot.

Issue #2630, opened 2018-10-25
Openly transphobic command in .rategirl command


minakitty:
I learned about this from another site, and was able to verify on both a server that I own and multiple servers that I'm an admin or member on that use this bot that it still exists.  It is specifically, this image: 
![screenshot_63](https://user-images.githubusercontent.com/5273418/47525848-1621d000-d853-11e8-8015-e1c060c13664.png)

I get that it's meant as a joke, likely referencing to "The Universal Hot vs Crazy Matrix" on YouTube, but the language of using "Tranny" and making them unrateable is highly offensive to many members of the transgender community.  Transgender people are, simply put, people.  Not some non-entity that deserve to be shamed and ridiculed.  And since people shouldn't be getting rated into that category anyway (because it's outside the chart), simply removing it from the base image file should be enough to clear up a number of issues

Potato1680:
Wha Wha Wha Wha Wha. Do you hear that? It's the sound of someone crying about something they don't like. Well, do I have the solution for you! If you don't like it, then don't use it! WOW!

Kaoticz:
If you don't like it, run `.sc ".rategirl" disable` on your Discord server.
The command will not be removed from the public bot, nor the image will be modified to suit your needs.

Alternatively, you can self-host your own Nadeko. You are free to modify whatever you wish on it.

JapanYoshi:
I disagree with both Kaoticz and Potato1680. You're putting this bot out into the public. You wouldn't put a command that shows you the N-word every time in such a bot, unless you support racism. It's the same with the T-slur and transphobia. It's simple human decency.

JapanYoshi:
![](https://encrypted-tbn0.gstatic.com/images?q=tbn%3AANd9GcSaN_DO2bk1UrNNWzqvowwM-DQzqha1buRxaHJ_J12YLYNdJmcT)

Kaoticz:
I wasn't arguing. I didn't say the command is not going to get removed/modified because I didn't agree with the presented point, I said the command is not getting removed/modified, period.

Hence why I presented a solution.
This is GitHub, not Tumblr.

JapanYoshi:
This isn't an argument. You may disagree with the point that your bot is needlessly transphobic. It is, and your opinion doesn't matter against the fact that your bot uses a pejorative.

[...]

PR #2632, opened 2018-10-26
Rategirl tweak

minakitty:
Now 38% less offensive.  Replaces tranny with trap, obligatory Star Wars reference, and updated image placeholder.  Potential resolution to #2630

### DIFF
--- a/NadekoBot.Core/Modules/Games/Games.cs
+++ b/NadekoBot.Core/Modules/Games/Games.cs
@@ -130,7 +130,8 @@ namespace NadekoBot.Modules.Games
                 hot = NextDouble(8, 10);
                 crazy = NextDouble(2, 3.99d);
                 advice = "You've met a girl she's above 8 hot, and not crazy at all (below 4)... totally cool?" +
-                         " You should be careful. That's a dude. You're talking to a tranny!";
+                         "It's a trap!  I mean, that's cool and all, just don't be spooked if their junk is bigger than yours." +
+                         "Just have a good time!";
             }
             else
             {
--- a/src/NadekoBot/data/images.json
+++ b/src/NadekoBot/data/images.json
@@ -27,7 +27,7 @@
     "https://nadeko-pictures.nyc3.digitaloceanspaces.com/other/dice/9.png"
   ],
   "Rategirl": {
-    "Matrix": "https://nadeko-pictures.nyc3.digitaloceanspaces.com/other/rategirl/matrix.png",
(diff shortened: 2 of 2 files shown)


minakitty:
As a note, the new matrix shouldn't stay on imgur.  Just happens to be that i wouldn't have access to the original hosting provider, presumably if the PR is accepted that change would be reverted and the image i put up would be moved to the cloud server Nadeko uses

GithubIsAwful:
Some people find "trap" offensive. The _real_ solution is to not change anything.

Anheledir:
The project owner is always free to close whatever request he doesn't like. It's his project.
So you have two rational possible solutions now:

1. Use your own self-hosted instance of the bot and replace the command and/or image with whatever you like. 
2. Don't use the bot at all and it's no longer your problem. You pointed out what in your personal opinion is a problem, those in charge don't agree, so there is nothing left for you to do. kthxbye.

minakitty:
Thank you so much for the insight Kowth.   Oh, wait ... never mind.

GithubIsAwful:
What did you add except a snarky comment toward the project lead?
Sod off.

minakitty:
While what I might have said had some snark, it was still a request for more information.

shikhir-arora:
This whole thing is so ridiculous. If you think that Kwoth is some anti-trans activist because he refuses to remove what amounts to a total meme (just look at it - does anyone take this seriously?!)

First, there's nothing wrong with this because it's a meme/joke. The **entire** thing. It's a Discord bot utility that is just for fun that can be disabled as pointed out or self-hosted yourself but you keep coming back and trying to change it because you're convinced that someone will be offended from this. I'm sorry to speak harshly, but if you're offended by this and can't see that it's nothing but a joke - a joke that fits only because you can assume this is rating a straight person. I don't know or care, I'm not transphobic at all, but this kind of stuff is infuriating. You have options: don't use the bot, disable the command, self host it, just don't keep insisting that you're right by changing the context. Also, you totally flipped the side here. "Just have fun!" - this ruins the point.

There's no one here that has anything against trans people, so please stop. No one is insinuating that a transgender person is "below human" just because of this joke, if you were able to understand the context of it all and how little this matters and how the people around here aren't generally rioting against the transgender population, then you'd see anyone should be able to laugh. I'm sorry but the world is way tougher out there, so if this is that offensive after all the discussion that took place in your last PR, I'm not sure if there's much anyone can do to make the world happy.

Starbors:
dang!!! this repo is bs

